PR #57, opened 2024-05-29
add sentry/raven support

Johnetordoff:
Adds Sentry support according to thier docs

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -1,6 +1,9 @@
 from pathlib import Path
 
 from app import env
+import os
+import raven
+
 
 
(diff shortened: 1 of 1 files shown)


aaxelb:
* pls check pre-commit checks
* why not [sentry-sdk](https://github.com/getsentry/sentry-python)? `raven` supports only up to py3.7; will be sheer luck if it works on 3.12
* how about adding the dep to `requirements/release.txt` and gracefully handle the missing import locally?

aaxelb:
also, mind sticking to the pattern where environment variables are only accessed in `app/env.py`?

PR #59, opened 2024-05-29
[ENG-5737] UUID primary keys

aaxelb:
* add `StrUUIDField` that is exactly the same as `UUIDField` but stringifies for use in python (so we don't have to change anything to handle UUID objects)
* update the base addon-service model to use `StrUUIDField` as primary key
  * note: can no longer use absence of `_instance.pk` to check whether the instance been saved -- instead, use [_instance._state.adding](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model._state)
* fix `do_box_test` (so could test it locally 👍 )

[ENG-5737]

[ENG-5737]: https://openscience.atlassian.net/browse/ENG-5737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwalz (approved):
Looks good. Let's figure out a merge strategy for these couple of tickets that are continuing to blow away Migratiion 0001

PR #60, opened 2024-05-29
consistent "display name" 

aaxelb:
- add `display_name` field to external services, authorized accounts, and configured addons
 - required for external service
 - if blank for account, fall back to external service
 - if blank for addon, fall back to account


aaxelb:
closing for #64

PR #61, opened 2024-06-03
[ENG-5738] encrypted credentials

aaxelb:
- rewrite `addon_service.credentials.models.ExternalCredentials` to encrypt credentials
- add `addon_service.credentials.encryption` with all the encrypt/decrypt logic
- new environment variables (see `app/env.py`)
  - in non-DEBUG environments, `GRAVYVALET_ENCRYPT_SECRET` is required


addon_service/credentials/encryption.py
```diff
@@ -0,0 +1,142 @@
...
+    "pls_encrypt_json",
+    "salt_factory",
+)
+
+
+# 32-byte key expected by cryptography.fernet.Fernet
+_KEY_BYTE_COUNT = 32
+
+# recommended len(salt) >= 16 bytes
+_SALT_BYTE_COUNT = settings.GRAVYVALET_SALT_BYTE_COUNT or 17
```
adlius, line 36: Should we also make note in the settings that `settings.GRAVYVALET_SALT_BYTE_COUNT` is recommended to be >= 16?

addon_service/credentials/encryption.py
```diff
@@ -0,0 +1,142 @@
...
+
+@dataclasses.dataclass(frozen=True)  # frozen for use as lru_cache key
+class KeyParameters:  # https://datatracker.ietf.org/doc/html/rfc7914#section-2
+    salt: bytes = dataclasses.field(default_factory=salt_factory)
+    # recommended scrypt_cost ("N") between 2^14 and 2^20
+    scrypt_cost: int = settings.GRAVYVALET_SCRYPT_COST or 2**17
+    # recommended scrypt_block_size ("r") = 8
+    scrypt_block_size: int = settings.GRAVYVALET_SCRYPT_BLOCK_SIZE or 8
+    # recommended scrypt_parallelization ("p") = 1
+    scrypt_parallelization: int = settings.GRAVYVALET_SCRYPT_PARALLELIZATION or 3
```
adlius, line 51: RFC 7914 recommends P=1. So should this be
```suggestion
    scrypt_parallelization: int = settings.GRAVYVALET_SCRYPT_PARALLELIZATION or 1
```

addon_service/credentials/encryption.py
```diff
@@ -0,0 +1,142 @@
...
+            and _cost_log2 == int(_cost_log2)
+            and _cost_log2 <= (128 * self.scrypt_block_size / 8)
+        )
+        # scrypt_parallelization "is a positive integer less than or equal to ((2^32-1) * 32) / (128 * r)"
+        assert self.scrypt_parallelization <= ((2**32 - 1) * 32) / (
+            128 * self.scrypt_block_size
+        )
+
+    def memory_required(self) -> int:
+        return self.scrypt_cost * self.scrypt_block_size * 129
```
adlius, line 68: Why `* 129`? I am sure there is a reason. I am simply curious.

addon_service/credentials/encryption.py
```diff
@@ -0,0 +1,142 @@
...
+        _fresh_params = stored_params
+    else:
+        # key defaults HAVE changed -- decrypt and re-encrypt
+        _decrypted = pls_decrypt_bytes(encrypted, stored_params)
+        _fresh_encrypted = pls_encrypt_bytes(_decrypted, _fresh_params)
+    return _fresh_encrypted, _fresh_params
+
+
+# deriving keys is expensive on purpose -- cache, but only in local memory
+@functools.lru_cache(maxsize=settings.GRAVYVALET_DERIVED_KEY_CACHE_SIZE)
```
adlius, line 109: This is a nitpick. But should `GRAVYVALET_DERIVED_KEY_CACHE_SIZE` be a calculated value instead of a constant? I assume the size of the key generated depends on the parameters used. So it might not always make sense to have a constant value for the cache, but a variable value based on `memory_required`?

adlius (commented):
Some questions.

addon_service/credentials/encryption.py
aaxelb: we should enforce that, probably... i've added `len(self.salt) >= 16` to what's validated in `KeyParameters.__post_init__` -- if/when we document the other parameters outside this file, should include that constraint

addon_service/credentials/encryption.py
aaxelb: hmm yeah, probably -- i bumped it arbitrarily after noticing django had [increased default parallelism to 5](https://code.djangoproject.com/ticket/35428), but on a closer look [the OWASP cheatsheet they cite](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt) recommends p=1 with our current default N=2^17 (tho it doesn't try to justify that recommendation 🤷 ) -- so yes, changed to 1

addon_service/credentials/encryption.py
aaxelb: i read somewhere that scrypt's approximate memory use is `N * r * 128` bytes, and fudged it up to 129 for `hashlib.scrypt` to stop throwing "memory limit exceeded" errors

...but now i've verified that estimate (it's a lower bound for any scrypt implementation not trying to play games recomputing intermediate values) and updated this function with explanatory comments and more empirically tested fudge

addon_service/credentials/encryption.py
aaxelb: the size of the derived key is a constant 32 bytes (as required by [cryptography.fernet.Fernet](https://cryptography.io/en/latest/fernet/#cryptography.fernet.Fernet) and set in `_KEY_BYTE_COUNT`) -- scrypt's parameters affect the amount of memory required for scrypt to run, but that memory can be (is?) freed as soon as it's done

addon_service/migrations/0001_initial.py
```diff
@@ -239,6 +218,36 @@ class Migration(migrations.Migration):
                 "verbose_name_plural": "External Storage Services",
             },
         ),
+        migrations.CreateModel(
```
mfraezz, line 221: Will this require a migration rollback / re-migrate when it hits staging?

app/env.py
```diff
@@ -31,3 +31,33 @@
...
+# 2. call `.rotate_encryption()` on every `ExternalCredentials` (perhaps via celery task
+#    `addon_service.credentials.tasks.schedule_encryption_rotation__celery`)
+# 3. remove the old secret from GRAVYVALET_ENCRYPT_SECRET_PRIORS
+GRAVYVALET_ENCRYPT_SECRET = os.environ.get("GRAVYVALET_ENCRYPT_SECRET")
+GRAVYVALET_ENCRYPT_SECRET_PRIORS = tuple(
+    filter(bool, os.environ.get("GRAVYVALET_ENCRYPT_SECRET_PRIORS", "").split(","))
+)
+# optional overrides for scrypt key derivation parameters (when unset or "0", use sensible defaults)
+# see https://datatracker.ietf.org/doc/html/rfc7914#section-2
+GRAVYVALET_SALT_BYTE_COUNT = int(os.environ.get("GRAVYVALET_SALT_BYTE_COUNT", 0))
```
mfraezz, line 52: Are any of these, aside from `GRAVYVALET_ENCRYPT_SECRET`, required to be set in the configs?

Also, why default to `0` here but override in the app code, rather than just defaulting to non-zero here?

mfraezz (approved):
A couple questions relating to deployment considerations, but otherwise LGTM :octocat:

addon_service/migrations/0001_initial.py
aaxelb: yes, tho staging is on `main` at the moment and there are other breaking migration changes in `develop` already -- i thiiink this PR is the last db-breaking change before it makes sense to freeze the initial migration

app/env.py
aaxelb: none required besides `GRAVYVALET_ENCRYPT_SECRET` (and that's optional too when `DEBUG`)

yeah, now looking again the `0` defaults here are weird and inconsistent... have added actual defaults here (and also replaced `SCRYPT_COST` with `SCRYPT_COST_LOG2` so you don't ever have to bother computing 2^18)

PR #62, opened 2024-06-03
[ENG-5697] admin interface (for setting up external services)

aaxelb:
- enable django admin (for ExternalStorageService and Oauth2ClientConfig)
- add formfield/widget for "linked" foreign keys -- allow viewing or (when null) adding a related object, but never list the related table


addon_service/common/known_imps.py
```diff
@@ -15,9 +15,11 @@
 
 
 __all__ = (
+    "AddonImpNumbers",
+    "KnownAddonImps",
```
aaxelb, line 19: added these to public module api since `AddonImpNumbers` now imported by admin

addon_service/external_storage_service/models.py
```diff
@@ -19,15 +19,18 @@ class ExternalStorageService(AddonsServiceBaseModel):
...
     int_credentials_format = models.IntegerField(
         null=False,
         validators=[validate_credentials_format],
+        verbose_name="Credentials format",
     )
     int_service_type = models.IntegerField(
         null=False,
         default=ServiceTypes.PUBLIC.value,
         validators=[validate_service_type],
+        verbose_name="Service type",
```
jwalz, line 33: Let's explicitly add `HYBRID = PUBLIC | HOSTED` to the ServiceTypes Flag so that it's selectable in Admin.

app/settings.py
```diff
@@ -40,15 +40,16 @@
...
+    "django.contrib.admin",
     "rest_framework",
     "rest_framework_json_api",
     "addon_service",
 ]
 
 if DEBUG:
     # run under ASGI locally:
-    INSTALLED_APPS.append("daphne")  # django's reference asgi server
+    INSTALLED_APPS.insert(0, "daphne")  # django's reference asgi server
```
jwalz, line 52: I know @mfraezz decided to go with daphne for the actual deployment. Maybe check and see if this can be moved out of the DEBUG?

aaxelb: i thiiink the only reason to mention `daphne` in app settings is to use it for `manage.py runserver` (an explicitly not-for-production dev tool), and we should avoid assuming a specific webserver (even when we know)

addon_service/configured_storage_addon/views.py
```diff
@@ -49,4 +49,5 @@ def get_wb_config(self, request, pk=None):
         addon = self.get_object()
         if addon.external_service.credentials_format is CredentialsFormats.OAUTH2:
             addon.base_account.refresh_oauth_access_token__blocking()
+        self.resource_name = "waterbutler-config"  # for the jsonapi resource type
```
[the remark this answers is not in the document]
aaxelb: oh hey, [it can](https://github.com/encode/django-rest-framework/blob/2da473c8c8e024e80c13a624782f1da6272812da/rest_framework/decorators.py#L163-L166)! but it [requires](https://github.com/encode/django-rest-framework/blob/2da473c8c8e024e80c13a624782f1da6272812da/rest_framework/viewsets.py#L94) those extra kwargs to already exist as attrs on the view -- which seemed fine, added `resource_name: str | None = None` to the viewset class and discovered jsonapi resource type [depends on an AttributeError](https://github.com/django-json-api/django-rest-framework-json-api/blob/027e44cc2219c9517cb6ae8ccf91e6eaf6ef01c8/rest_framework_json_api/utils.py#L52)... could try switching to defining all `resource_name`s on the viewsets instead of on the models (help keep api concerns out of the data layer) but iirc something drf-jsonapi got in the way of that

addon_service/external_storage_service/models.py
aaxelb: turns out iterating thru an `enum.Flags` omits the aliases -- had to update to use `__members__`

adlius (approved):

addon_service/configured_storage_addon/views.py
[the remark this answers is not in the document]
jwalz: The joys of opinionated frameworks

PR #64, opened 2024-06-06
[ENG-5460] converge with frontend

aaxelb:
- add `display_name` field to external services, authorized accounts, and configured addons
    - required for external service
    - if blank for addon, fall back to account
    - if blank for account, fall back to external service
- set jsonapi type for `/configured-storage-accounts/{pk}/waterbutler-config/` (now has `"type": "waterbutler-config"`)
- ~admin interface (merged from #62 )~
- updates to [converge with frontend](https://github.com/CenterForOpenScience/ember-osf-web/pull/2234)
  - when `/v1/user-references?filter[user_uri]=...` or `/v2/resource-references?filter[resource_uri]=...` match no results, respond success with `"data": []` instead of 404 error
  - allow empty `"credentials": {}` on account creation, consider same as when omitted
  - allow blank `root_folder` for addons
  - stop adding trailing slashes
  - create `UserReference` to current user on successful auth

[ENG-5460]

[ENG-5460]: https://openscience.atlassian.net/browse/ENG-5460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/addon_service/addon_operation_invocation/views.py
+++ b/addon_service/addon_operation_invocation/views.py
@@ -2,10 +2,13 @@ from addon_service.common.permissions import (
     IsAuthenticated,
     SessionUserIsOwner,
     SessionUserMayAccessInvocation,
+    SessionUserMayPerformInvocation,
 )
 from addon_service.common.viewsets import RetrieveWriteViewSet
+from addon_toolkit import AddonOperationType
 
 from .models import AddonOperationInvocation
+from .perform import perform_invocation__blocking
 from .serializers import AddonOperationInvocationSerializer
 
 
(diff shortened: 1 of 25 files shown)


addon_service/addon_operation_invocation/views.py
```diff
@@ -20,10 +23,23 @@ def get_permissions(self):
...
+    def perform_create(self, serializer):
+        super().perform_create(serializer)
+        # after creating the AddonOperationInvocation, look into invoking it
+        _invocation = serializer.instance
+        _operation_type = _invocation.operation.operation_type
+        match _operation_type:
+            case AddonOperationType.REDIRECT | AddonOperationType.IMMEDIATE:
+                perform_invocation__blocking(_invocation)
+            case AddonOperationType.EVENTUAL:
+                raise NotImplementedError("TODO: enqueue task")
```
adlius, line 43: Does this mean none of our current addon ops are of `EVENTUAL` type?

adlius (approved):
Looks good. Just one small question.

addon_service/addon_operation_invocation/views.py
aaxelb: correct -- the only operations we need at this phase are `IMMEDIATE` (for read-only filetree browsing)

`EVENTUAL` operations may (eventually) include filetree copy/move/delete operations (still handled by waterbutler) and archiving (still handled by osf's archiver and/or pigeon), but we're deferring the complexity of "eventual" state handling until we actually want to use it

addon_service/user_reference/models.py
```diff
@@ -16,6 +17,12 @@ def configured_storage_addons(self):
             base_account__account_owner=self,
         )
 
+    @property
+    def configured_resources(self):
+        return ResourceReference.objects.filter(
+            configured_storage_addons__base_account__account_owner=self,
```
jwalz, line 23: will this get duplicates if one resource has multiple configured addons by the same user? If so, might need to be something like:
```
ResourceReference.objects.annotate(
  has_addon_configured_by_user=Exists(
    ConfiguredStorageAddon.objects.all.filter(
      confiugred_resource_id=OuterRef('id'),
      base_account__account_owner=self
    )
).filter(has_addons_configured_by_user=True)
```

jwalz (commented):
One open question, but generally lgtm. Thanks!

addon_service/user_reference/models.py
aaxelb: oh, probably yes duplicates -- how do you think the subquery compares to thowing a `.distinct()` on there?

jwalz: My understanding is that `distinct` performance is usually pretty poor.

aaxelb: feels like a case (with probably usually low row counts and multiplicity) where `SELECT DISTINCT` would do fine enough, but haven't benchmarked and am fine with the subquery

PR #65, opened 2024-06-12
[ENG-5725] support hmac-authenticated requests from osf

aaxelb:
- port some updates to `addon_service.common.hmac` from [osf.external.gravy_valet.auth_helpers](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/external/gravy_valet/auth_helpers.py)
- add specific exceptions for hmac situations
- when receiving an api request from osf, allow authenticating and authorizing with hmac-signed headers:
  - `X-Requesting-User-URI` for current user
  - `X-Requested-Resource-URI` for a single relevant resource
  - `X-Requested-Resource-Permissions` for the current user's osf permissions on the relevant resource
- rename setting `DEFAULT_HMAC_KEY` to `OSF_HMAC_KEY`; update hmac utils to require a key (easing futures with multiple hmac keys in play)

[ENG-5725]

[ENG-5725]: https://openscience.atlassian.net/browse/ENG-5725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/addon_service/common/hmac.py
+++ b/addon_service/common/hmac.py
@@ -19,9 +19,14 @@ __all__ = (
     "TIMESTAMP_HEADER",
 )
 
-_AUTH_HEADER_PREFIX = "HMAC-"
+# this is but one way to hmac-sign an http request -- aligns with how osf sends them:
+# https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/external/gravy_valet/auth_helpers.py
+
+_AUTH_HEADER_SCHEME = "HMAC-SHA256"
 _AUTH_HEADER_REGEX = re.compile(
-    r"^HMAC-SHA256 SignedHeaders=(?P<headers>[\w;-]*)&Signature=(?P<signature>[^\W_]*$)"
+    "^"
+    + re.escape(_AUTH_HEADER_SCHEME)
+    + r" SignedHeaders=(?P<headers>[\w;-]*)&Signature=(?P<signature>[^\W_]*$)"
 )
 TIMESTAMP_HEADER = "X-Authorization-Timestamp"
 CONTENT_HASH_HEADER = "X-Content-SHA256"
@@ -45,9 +50,7 @@ def make_signed_headers(
         message="\n".join(signed_string_segments), hmac_key=hmac_key
     )
     signature_header_fields = ";".join(signed_headers.keys())
-    auth_header_value = (
-        f"HMAC-SHA256 SignedHeaders={signature_header_fields}&Signature={signature}"
-    )
+    auth_header_value = f"{_AUTH_HEADER_SCHEME} SignedHeaders={signature_header_fields}&Signature={signature}"
     return dict(
         **signed_headers,
         Authorization=auth_header_value,
@@ -59,7 +62,7 @@ def validate_signed_headers(
     request: HttpRequest, hmac_key: str, expiration_seconds: int | None = None
 ) -> dict[str, str]:
     _authorization = request.headers.get("Authorization", "")
-    if not _authorization.startswith(_AUTH_HEADER_PREFIX):
+    if not _authorization.startswith(_AUTH_HEADER_SCHEME):
         raise NotUsingHmac
     match = _AUTH_HEADER_REGEX.match(_authorization)
     if not match:
--- a/addon_service/common/hmac.py
+++ b/addon_service/common/hmac.py
@@ -1,62 +1,49 @@
 import base64
+import functools
 import hashlib
 import hmac
 import re
-import urllib
+import urllib.parse
 from datetime import (
     UTC,
     datetime,
+    timedelta,
 )
 
-from django.conf import settings
+from django.http import HttpRequest
 
 
+__all__ = (
+    "make_signed_headers",
+    "validate_signed_headers",
+    "TIMESTAMP_HEADER",
+)
+
+_AUTH_HEADER_PREFIX = "HMAC-"
 _AUTH_HEADER_REGEX = re.compile(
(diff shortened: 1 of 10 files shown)


addon_service/common/hmac.py
```diff
@@ -1,62 +1,49 @@
...
-    signed_segments = [segment for segment in signed_segments if segment]
-    signed_headers = {"X-Authorization-Timestamp": auth_timestamp}
-    if content_hash:
-        signed_headers["X-Content-SHA256"] = content_hash
-    return signed_segments, signed_headers
+###
+# public
 
 
 def make_signed_headers(
```
adlius, line 34: It seems that this method is only invoked in tests to mock an HMAC-authenticated requests from OSF. If so we should probably make a comment, unless there are plans to have GV make HMAC requests to other services.

addon_service/common/hmac.py
```diff
@@ -67,37 +54,145 @@ def make_signed_headers(
...
-
-
-def validate_signed_headers(request, hmac_key=None):
-    match = _AUTH_HEADER_REGEX.match(request.headers.get("Authorization", ""))
+@functools.lru_cache
+def validate_signed_headers(
+    request: HttpRequest, hmac_key: str, expiration_seconds: int | None = None
+) -> dict[str, str]:
+    _authorization = request.headers.get("Authorization", "")
+    if not _authorization.startswith(_AUTH_HEADER_PREFIX):
```
adlius, line 62: Are all HMAC-authenticated requests guaranteed to have an `Authorization` field in the header that starts with `HMAC-`? I have been looking online and there does not seem to be a standard. In which case it is an agreement between the sender (OSF) and the receiver (GV). If that is the case, does it make sense to have `_AUTH_HEADER_PREFIX` to be something imported from local env vars shared by both OSF and GV?

addon_service/common/osf.py
```diff
@@ -17,6 +19,11 @@
     "has_osf_permission_on_resource",
 )
 
+_OSF_HMAC_USER_HEADER = "X-Requesting-User-URI"
+_OSF_HMAC_RESOURCE_HEADER = "X-Requested-Resource-URI"
+_OSF_HMAC_PERMISSIONS_HEADER = "X-Requested-Resource-Permissions"
+_OSF_HMAC_PERMISSIONS_DELIMITER = ";"  # see https://github.com/CenterForOpenScience/osf.io/blob/00a17ecdb666c07245a520de3ccd0f743822573b/osf/external/gravy_valet/auth_helpers.py#L66
+
```
adlius, line 26: Does it make sense to have these come from a common set of local env vars shared by OSF and GV? Or am I overreacting?

adlius (commented):
I only understands HMAC in principle so I have some dumb questions.

addon_service/common/hmac.py
aaxelb: i don't think we have specific plans for GV to send hmac-verified requests to other services, but i think having this function here seems better than moving it to a test helper -- more easily allows possible futures where gv does that or where we consolidate hmac utils as part of a shared dependency and use the same code on both ends of the hmac requests

addon_service/common/hmac.py
aaxelb: oh good point, i assumed wrong about the `HMAC-` prefix without checking for a standard -- correct we need only agreement with OSF (another reason shared code across services would be helpful), will update the errors accordingly (and perhaps add environment variables, yeah)

addon_service/common/hmac.py
```diff
@@ -67,37 +54,145 @@ def make_signed_headers(
...
-    signed_segments.extend(
-        [str(request.headers[signed_header]) for signed_header in signed_headers]
-    )
-    return "\n".join([segment for segment in signed_segments if segment])
-
-
-def validate_signed_headers(request, hmac_key=None):
-    match = _AUTH_HEADER_REGEX.match(request.headers.get("Authorization", ""))
+@functools.lru_cache
+def validate_signed_headers(
```
jwalz, line 61: Nit, because it feels unusual to have a `validate...` function return a result (and be called many times).
```suggestion
def get_hmac_signed_headers(
```

Option b (for better?):
Leave `validate_signed_headers` without a return value but still cache the result. Add a separate `get_signed_headers` function that checks `validate_signed_headers` and returns only the headers that are part of the hmac signature.

app/env.py
```diff
@@ -22,6 +22,7 @@
 
 SECRET_KEY = os.environ.get("SECRET_KEY")
 OSF_HMAC_KEY = os.environ.get("OSF_HMAC_KEY")
+OSF_HMAC_EXPIRATION_SECONDS = int(os.environ.get("OSF_HMAC_KEY", 110))
```
jwalz, line 25:
```suggestion
OSF_HMAC_EXPIRATION_SECONDS = int(os.environ.get(SOMETHING_ELSE, 110))
```

addon_service/common/hmac.py
aaxelb: how about rename to `get_signed_headers` (and keep cache), then add `validate_signed_request` that calls `get_signed_headers` and returns nothing? (this is what i have done, open to thoughts)

addon_service/common/osf.py
aaxelb: not a bad idea, and i agree with the urge to genericize... but there's enough logic that has to be shared i'm not sure we gain anything but complexity by moving these to more env vars (at least until we have the code shared on both sides)

addon_service/common/hmac.py
jwalz: works for me

jwalz (approved):

adlius (approved):

PR #66, opened 2024-06-13
[ENG-5804] - Mariia's GV local setup

ly-mariia:
- changed the order in the Dockerfile, docker didn't work because manage.py was run before installing requirements;
- added in README.md makemigrations command;
- changed the python3 command to python since we have one python;


addon_service/migrations/0002_alter_externalstorageservice_int_addon_imp_and_more.py
```diff
@@ -0,0 +1,59 @@
...
+            name="oauth2_client_config",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="external_storage_services",
+                to="addon_service.oauth2clientconfig",
+            ),
+        ),
+    ]
```
adlius, line 59: Looks like we made some model changes without `makemigrations`. @jwalz @aaxelb Do we want to add a new migration file or do we want to fold it into `0001_initial`?

adlius (approved):
Thanks for updating the Dockerfile and the readme. Looks good.

addon_service/migrations/0002_alter_externalstorageservice_int_addon_imp_and_more.py
aaxelb: whoops, yeah i missed that

we are still updating `0001_initial` in place (`rm addon_service/migrations/000* && python manage.py makemigrations`), at least until we have external services set up on staging

Dockerfile
```diff
@@ -6,15 +6,16 @@ RUN apt-get update && apt-get install -y libpq-dev
...
 # END gv-base
 
 # BEGIN gv-deploy
 FROM gv-base as gv-deploy
 # only non-dev dependencies:
 RUN pip3 install --no-cache-dir -r requirements/requirements.txt
 # note: no CMD in gv-deploy -- depends on deployment
+
+RUN python manage.py collectstatic --noinput
+EXPOSE 8000
```
aaxelb, line 18: another whoops -- good catch, thx

aaxelb (commented):
thanks for catching these mistakes! i've opened a pr with fixes at #68

aaxelb:
fixes added with #68 -- thanks!

PR #67, opened 2024-06-14
Check whether test is running

adlius:
Currently, if we run tests locally, it is asking us to add 'osf' to the list of databases for tests. We have managed to avoid this by only appending the "osf" database configs when the `env.OSFDB_HOST` is set. This works for CI test runs but the same problem occurs locally when we try to run tests.

This PR tries to solve this problem by checking `sys.argv` to see if we are running tests. It works, but is this a hacky solution? Need some inputs/suggestions. @jwalz @aaxelb 

H/T @ly-mariia for discovering the problem.

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -1,3 +1,4 @@
+import sys
 from pathlib import Path
 
 from app import env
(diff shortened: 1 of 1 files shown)


app/settings.py
```diff
@@ -103,7 +104,7 @@
     },
 }
 
-if env.OSFDB_HOST:
+if env.OSFDB_HOST and not sys.argv[1:2] == ["test"]:
```
aaxelb, line 107: yeah, this feels brittle... i tried an alternate fix in #68 that seemed to work (disabling django's automatic transaction handling), but then ran into another osfdb error from `makemigrations` (because my local osf postgres is still 9.x...) and decided to remove the default `OSFDB_HOST` until we have credentials migration logic using it

aaxelb:
#68 merged with an alternate workaround

PR #69, opened 2024-06-17
Fix/cors middleware

opaduchak:
Fixed not being able to call GV from frontend due to missing CORS headers


aaxelb (commented):
looks good! would you mind rebasing your changes onto `develop`? (i've now merged `assorted-betterment`, and you shouldn't need the migration changes)

addon_service/migrations/0002_alter_externalstorageservice_int_addon_imp_and_more.py
aaxelb, line 1: you should be able to delete this (the `verbose_name` changes are now reflected in `0001_initial`

aaxelb (changes requested):
remove the changes to migrations and it's good to go, thanks!

app/env.py
```diff
@@ -31,3 +31,4 @@
 
 # comma-separated list:
 ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "").split(",")
+CORS_ALLOWED_ORIGINS = os.environ.get("CORS_ALLOWED_ORIGINS", "").split(",")
```
aaxelb, line 34: when CORS_ALLOWED_ORIGINS isn't set it defaults to `[""]`, which [causes an error](https://github.com/CenterForOpenScience/gravyvalet/actions/runs/9598391564/job/26479890621?pr=69#step:8:23) 
```suggestion
CORS_ALLOWED_ORIGINS = tuple(filter(bool, os.environ.get("CORS_ALLOWED_ORIGINS", "").split(",")))
```
(i guess ALLOWED_HOSTS is not validated as strictly, but could do with the same fix)

aaxelb (changes requested):
small bug from the CI run

app/env.py
opaduchak: done

addon_service/migrations/0002_alter_externalstorageservice_int_addon_imp_and_more.py
opaduchak: done

aaxelb (approved):
👍

opaduchak:
So, can I merge this now @aaxelb?

PR #73, opened 2024-06-27
[ENG-5730][ENG-5731] gravyvalet code docs

aaxelb:
- use `pdoc` to generate code docs from docstrings
- update `__all__` usage (since it tells pdoc what (not) to include), especially replace `__all__ = ()` with docstring
- add `gravyvalet_code_docs/build.py` to generate docs; run from Dockerfile for non-dev builds
- improve docstrings and readmes (readsme?)
- generated docs currently available at `/docs` (redirects to `/static/gravyvalet_code_docs/index.html`, but let's use `/docs` in links)
- see built docs at https://addons.staging.osf.io/docs

[ENG-5730]
[ENG-5731]

[ENG-5730]: https://openscience.atlassian.net/browse/ENG-5730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-5731]: https://openscience.atlassian.net/browse/ENG-5731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/addon_service/common/network.py
+++ b/addon_service/common/network.py
@@ -55,6 +55,8 @@ class _AiohttpResponseInfo(HttpResponseInfo):
 
 
 class GravyvaletHttpRequestor(HttpRequestor):
+    """an `HttpRequestor` implementation using aiohttp"""
+
     # abstract property from HttpRequestor:
     response_info_cls = _AiohttpResponseInfo
 
--- a/addon_service/common/network.py
+++ b/addon_service/common/network.py
@@ -12,15 +12,18 @@ from urllib.parse import (
 import aiohttp
 from asgiref.sync import sync_to_async
 
-from addon_service import models as db
 from addon_service.common import exceptions
 from addon_service.common.credentials_formats import CredentialsFormats
-from addon_toolkit.constrained_network import (
+from addon_toolkit.constrained_network.http import (
     HttpRequestInfo,
     HttpRequestor,
     HttpResponseInfo,
 )
 from addon_toolkit.iri_utils import Multidict
+
+
+if typing.TYPE_CHECKING:
+    from addon_service.models import AuthorizedStorageAccount
 
 
 __all__ = ("GravyvaletHttpRequestor",)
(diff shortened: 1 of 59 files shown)


adlius (approved):

jwalz:
A thing that I'm noticing: the docs for Django/DRF related things are inheriting a lot of fundamentally un-useful descriptors (for instance, every single field says "A wrapper for a deferred-loading field. When the value is read from this object the first time, the query is executed."). Any way to omit those and/or to surface helpful comments of our own articulating the purpose of each field?

addon_service/common/network.py
```diff
@@ -60,7 +63,7 @@ def __init__(
         *,
         client_session: aiohttp.ClientSession,
         prefix_url: str,
-        account: db.AuthorizedStorageAccount,
+        account: "AuthorizedStorageAccount",
```
jwalz, line 66: I *think* that newer versions of python don't require the quotes -- though it may still require a `from __future__ import annotations` to omit them?

addon_service/common/network.py
```diff
@@ -12,17 +12,20 @@
...
+from addon_toolkit.constrained_network.http import (
     HttpRequestInfo,
     HttpRequestor,
     HttpResponseInfo,
 )
 from addon_toolkit.iri_utils import Multidict
 
 
+if typing.TYPE_CHECKING:
+    from addon_service.models import AuthorizedStorageAccount
```
jwalz, line 26: 🎉 🎉 🎉

aaxelb:
> A thing that I'm noticing: the docs for Django/DRF related things are inheriting a lot of fundamentally un-useful descriptors (for instance, every single field says "A wrapper for a deferred-loading field. When the value is read from this object the first time, the query is executed."). Any way to omit those and/or to surface helpful comments of our own articulating the purpose of each field?

yeah that's no good... could either exclude those modules or add explicit docstrings to every field (tho nonstandard, pdoc does support a docstring on the line after each class attribute)

excluding the modules is probably much easier, so i ask: is there value in these kinds of code docs for models, viewsets, and serializers? or would those be better replaced by json-ful api docs?

addon_service/common/network.py
aaxelb: ohh, for some reason i thought `from __future__ import annotations` was for use in older versions to get current behavior (the _past_'s future), but yeah that's exactly what should do here

PR #74, opened 2024-06-27
fixups for deployment

aaxelb:
- stop raising error on missing GRAVYVALET_ENCRYPT_SECRET in app.settings (or, if DEBUG, defaulting to a bad secret) -- allow it to be `None`
- give a helpful error when trying to encrypt/decrypt without GRAVYVALET_ENCRYPT_SECRET
- set a fake GRAVYVALET_ENCRYPT_SECRET in docker-compose.yml for local dev use with made-up data
- use fake GRAVYVALET_ENCRYPT_SECRET in tests
- set STATIC_ROOT (for `manage.py collectstatic`)
- add `rotate_encryption` management command (schedules a celery task)
- add sentry_sdk


addon_service/authorized_storage_account/serializers.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, *args, **kwargs):
         view_name=view_names.detail_view(RESOURCE_TYPE), required=False
     )
     display_name = serializers.CharField(
-        allow_blank=True, required=False, max_length=256
+        allow_blank=True, allow_null=True, required=False, max_length=256
```
jwalz, line 42: nit: is there a reason for this to be nullable vs having a default `""`? i.e. do we need to distinguish between the differing empty values?

aaxelb: i encountered the frontend sending `display_name: null`, which i think would error without `allow_null=True` (tho could also talk about a frontend fix)

jwalz: I missed that this was in the serializer, so I retract my complaint. That said, I *think* that `required=False` will just not load it into `validated_data` if it's `None`? I know that `required=False` won't include a `None` value in the serialized output, but not sure if that behavior is bi-directional...

addon_service/authorized_storage_account/serializers.py
```diff
@@ -112,6 +112,26 @@ def create(self, validated_data):
 
         return authorized_account
 
+    def update(self, instance, validated_data):
```
jwalz, line 115: Do we want to do the work to complain if we're given values to update that aren't supported?

jwalz: (this is an actual question, not midwestern passive aggression)

aaxelb: hmm could do... tho unfortunately, ember-data doesn't (iirc) have a sense of "read-only" or "write-once-on-create" attributes, so all fields are included on every PATCH -- could add frontend work to avoid that, but seems easier to ignore non-updatable fields (leaning toward permissiveness in what we accept)

jwalz: Yeah, we would need to compare all non-updatable values to what is currently set if we wanted to have explicitly "non-updatable" fields.

addon_service/authorized_storage_account/serializers.py
aaxelb: i think it's [different on the way in](https://github.com/encode/django-rest-framework/blob/2da473c8c8e024e80c13a624782f1da6272812da/rest_framework/fields.py#L515-L524) -- `required=False` allows missing keys, but `None` is rejected unless `allow_none=True`

PR #78, opened 2024-07-05
[ENG-5769] Oauth 1.0a integration

opaduchak:
Implementation of Oauth1.0a for GravyValet
[ENG-5769](https://openscience.atlassian.net/browse/ENG-5769)

[ENG-5769]: https://openscience.atlassian.net/browse/ENG-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


addon_toolkit/credentials.py
```diff
@@ -27,6 +27,29 @@ class AccessKeySecretKeyCredentials(Credentials):
     secret_key: str
 
 
+@dataclasses.dataclass(frozen=True, slots=True)
+class OAuth1TokenCredentials:
```
adlius, line 31: I think for OAuth1 we should still use `AccessTokenCredentials`. The idea of the `AccessTokenCredentials` dataclass is to implement the `iter_headers` method so that `GravyvaletHttpRequestors` can automatically append the appropriate auth headers for the oauth requests.

addon_service/authorized_storage_account/models.py
```diff
@@ -76,6 +76,12 @@ class AuthorizedStorageAccount(AddonsServiceBaseModel):
         related_name="authorized_storage_accounts",
     )
 
+    is_oauth1_ready = models.BooleanField(
+        "addon_service.OAuth2TokenMetadata",
+        null=True,
+        blank=True,
+    )
+
```
adlius, line 84: I am not quite sure whether this boolean field is necessary.

addon_service/tests/_factories.py
```diff
@@ -89,6 +99,27 @@ def _create(
...
+        cls,
+        model_class,
+        credentials_format=CredentialsFormats.OAUTH1A,
+        service_type=ServiceTypes.PUBLIC,
+        *args,
+        **kwargs,
+    ):
+        return super()._create(
+            model_class, credentials_format, service_type, *args, **kwargs
+        )
```
jwalz, line 120: We never shoul had `oauth_client_config` as a sub-factory, considering it isnt' relevant and shouldn't be set for non-oauth(2) services.

Would recommend instead just updating the existing `ExternalStorageServiceFactory._create` function to
1) Accept `oauth1_client_config` and `oauth2_client_config` as parameters
2) Use the `credentials_format` to construct the appropriate form of client config using the appropriate factory if one wasn't passed
3) Ignore or error any irrelevant `client_config` that was passed.

jwalz: Note, this is obviously simpler if we decide to collapse down to one `OAuthConfig`.

addon_service/oauth1/utils.py
```diff
@@ -0,0 +1,148 @@
...
+            return OAuth1TokenCredentials.from_dict(result_dict)
+
+
+def _construct_params(params_to_encode: dict) -> str:
+    return ",".join(
+        [f'{key}="{params_to_encode[key]}"' for key in sorted([*params_to_encode])]
+    )
+
+
+def construct_signed_headers(
```
jwalz, line 83: Do any of our existing [`hmac` utils ](https://github.com/CenterForOpenScience/gravyvalet/blob/develop/addon_service/common/hmac.py)help here? And can they be refactored to be more general to support this case?

addon_service/authorized_storage_account/models.py
```diff
@@ -76,6 +76,12 @@ class AuthorizedStorageAccount(AddonsServiceBaseModel):
         related_name="authorized_storage_accounts",
     )
 
+    is_oauth1_ready = models.BooleanField(
+        "addon_service.OAuth2TokenMetadata",
```
jwalz, line 80:
```suggestion
```

addon_service/authorized_storage_account/models.py
```diff
@@ -76,6 +76,12 @@ class AuthorizedStorageAccount(AddonsServiceBaseModel):
         related_name="authorized_storage_accounts",
     )
 
+    is_oauth1_ready = models.BooleanField(
```
jwalz, line 79: looks like we need this because we lack another way to distinguish between temporary and permanent credentials? would it be appropriate to add a field for tracking this to the OAuth1TokenCredentials?

addon_service/authorized_storage_account/serializers.py
```diff
@@ -95,8 +95,14 @@ def create(self, validated_data):
             authorized_account.initiate_oauth2_flow(
                 validated_data.get("authorized_scopes")
             )
+        elif external_service.credentials_format is CredentialsFormats.OAUTH1A:
+            authorized_account.initiate_oauth1_flow()
+            self.context["request"].session[
+                "oauth1a_account_id"
+            ] = authorized_account.pk
```
adlius, line 102: Do we have to store the `authorized_account.pk` in the session? Is there anything in the payload of the request to the callback url that can be used to identify which authorized account this is?

addon_service/oauth1/utils.py
```diff
@@ -0,0 +1,148 @@
...
+        except ContentTypeError:
+            raw_result = parse_qs(await _token_response.text())
+            result_dict = {key: value[0] for key, value in raw_result.items() if value}
+            return OAuth1TokenCredentials.from_dict(result_dict)
+
+
+def _construct_params(params_to_encode: dict) -> str:
+    return ",".join(
+        [f'{key}="{params_to_encode[key]}"' for key in sorted([*params_to_encode])]
+    )
```
jwalz, line 80: sorted(dict.items()) actually works the way you hope it would, and no need to pass an explicit list to `join` instead of just the expression.

As such, this can be
```suggestion
    return ",".join(
        f"{key}={value}" for key, value in sorted(params_to_encode.items())
    )

File: addon_service/oauth1/views.py [hunk 9]
```diff
@@ -0,0 +1,42 @@
...
+    account = AuthorizedStorageAccount.objects.get(pk=pk)
+
+    oauth1_client_config = account.external_service.oauth1_client_config
+    final_credentials, other_info = async_to_sync(get_access_token)(
+        oauth1_client_config.access_token_url,
+        oauth1_client_config.client_key,
+        oauth1_client_config.client_secret,
+        oauth_token,
+        account.credentials.oauth_token_secret,
+        oauth_verifier,
```
jwalz, line 28: nit: kwargs preferred to positional args, especially with this many arguments

adlius (commented):
First round. Some questions.

addon_service/tests/_helpers.py
```diff
@@ -159,6 +173,79 @@ async def _route_post(self, url, *args, **kwargs):
...
+
+    def set_internal_client(self, client):
+        """Attach a DRF APIClient for making requests internally"""
+        self._internal_client = client
+
+    @contextlib.asynccontextmanager
+    async def amocking(self):
+        client_session = await get_singleton_client_session()
+        with (
+            patch.object(client_session, "get", new=self._route_get),
```
jwalz, line 206: Not seeing a defintion for `self._route_get`

addon_imps/citations/zotero_org.py
```diff
@@ -0,0 +1,5 @@
+from addon_toolkit.interfaces.storage import StorageAddonImp
+
+
+class ZoteroOrgCitationImp(StorageAddonImp):
```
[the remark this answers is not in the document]
opaduchak: this is a crutch because we still don't have citation support inside `gravyvalet`. I tried constructing dummy `CitationAddonImp`, but this has created more problems than it solved

addon_service/common/credentials_formats.py
```diff
@@ -1,20 +1,27 @@
...
     PERSONAL_ACCESS_TOKEN = 4
+    OAUTH1A = 5
 
     @property
     def dataclass(self):
         match self:
             case CredentialsFormats.OAUTH2:
                 return credentials.AccessTokenCredentials
+            case CredentialsFormats.OAUTH1A:
+                return credentials.OAuth1TokenCredentials
```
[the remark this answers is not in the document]
opaduchak: There is `oauth_token` and `oauth_token_secret`, which in their nature and usage are drastically different from OAuth2 access and refresh tokens

jwalz (commented):
Thanks! This is a great start. I'll probably dig a bit more into the testing stuff to see if there's an easy way to consolidate strategies.

addon_service/oauth1/models.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Model for storing attributes that are required for managing
+    OAuth1 credentials exchanges with an ExternalService on behalf
+    of a registered client (e.g. the OSF)
+    """
+
+    # URI that allows to obtain temporary request token to proceed with user auth
+    request_token_url = models.URLField(null=False)
+    # URI to which user will be redirected to authenticate
+    auth_url = models.URLField(null=False)
```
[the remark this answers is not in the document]
opaduchak: I dunno, If you say so, maybe it is

addon_service/oauth1/models.py
```diff
@@ -0,0 +1,31 @@
+from django.db import models
+
+from addon_service.common.base_model import AddonsServiceBaseModel
+
+
+class OAuth1ClientConfig(AddonsServiceBaseModel):
```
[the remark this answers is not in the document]
opaduchak: `auth_url` is `auth_callback_url`, I'll change it if it's confusing

addon_toolkit/credentials.py
jwalz: Depends on if OAuth1 requests require the secret to be sent along with the token. If so, it's a different credentials shape. Otherwise, yes, we should keep using AccessTokenCredentials and create an OAuth1 equivalent to `OAuth2TokenMetadata` for tracking other details

opaduchak: Well, auth headers are different for OAuth1.0a and OAuth2, and even more: they are different for Zotero

addon_service/oauth1/models.py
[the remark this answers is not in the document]
jwalz: For the record: we're storing the `auth_callback_url` explicitly in case there are services that won't allow us to update the list of callbacks without invalidating our client and/or old credentials. That way, if necessary, we can encode this as `https://[environment.]osf.io/oauth/callback/[provider]` -- where we already have things set up to forward requests to GV if appropriate.

addon_service/credentials/serializers.py
```diff
@@ -17,7 +18,7 @@ def __init__(self, write_only=True, required=False, *args, **kwargs):
         super().__init__(write_only=write_only, required=required)
 
     def to_internal_value(self, data):
-        if not data:
+        if not data or not any(data.values()):
```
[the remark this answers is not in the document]
opaduchak: Ok, then I'll revert this

addon_service/oauth1/utils.py
opaduchak: They don't help, because OAuth1.0a constructs it's signature base in different way, and it embeds this signature in different way. As a result I doubt that refactoring them would pay off. I know DRY is good, but I'd consider this if we had anything else which requires similar hmac approach

jwalz: :+1:

addon_service/authorized_storage_account/models.py
opaduchak: You are right, this is the purpose for this field. Your approach is also possible

addon_service/oauth1/models.py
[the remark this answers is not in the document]
jwalz: In this case, it' guaranteed to be a url. I think the only issue right now is that we don't really have a standard for when we use `url` vs `uri`, which is something we should settle on as part of code cleanup and documentation tasks.

addon_service/authorized_storage_account/models.py
opaduchak: It is necessary as it allows to discern temporary Request token credentials from final credentials

addon_service/oauth1/views.py
opaduchak: Yep, but also they are kinda self-explainatory

addon_service/oauth1/models.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # URI that allows to obtain temporary request token to proceed with user auth
+    request_token_url = models.URLField(null=False)
+    # URI to which user will be redirected to authenticate
+    auth_url = models.URLField(null=False)
+    # URI to obtain access token
+    access_token_url = models.URLField(null=False)
+
+    client_key = models.CharField(null=True)
+    client_secret = models.CharField(null=True)
```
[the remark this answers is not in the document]
opaduchak: 1. Yes
2. Yes
Well, purpose for client_key is the same, while client_secret is used to sign requests (in combination with oauth_token secret if we have it)

addon_service/oauth1/utils.py
opaduchak: Didn't apply your suggestion, because you missed " around value

addon_service/tests/_helpers.py
```diff
@@ -159,6 +173,70 @@ async def _route_post(self, url, *args, **kwargs):
             raise RuntimeError(f"Received unrecognized endpoint {url}")
 
 
+@dataclasses.dataclass
+class MockServiceProvider:
```
aaxelb, line 177: naming nit: since the code is specific to oauth1, how about `MockOauth1ServiceProvider` or `MockOauth1Server`?

addon_service/oauth1/utils.py
```diff
@@ -0,0 +1,148 @@
...
+            return OAuth1TokenCredentials.from_dict(await _token_response.json())
+        except ContentTypeError:
+            raw_result = parse_qs(await _token_response.text())
+            result_dict = {key: value[0] for key, value in raw_result.items() if value}
+            return OAuth1TokenCredentials.from_dict(result_dict)
+
+
+def _construct_params(params_to_encode: dict) -> str:
+    return ",".join(
+        [f'{key}="{value}"' for key, value in sorted(params_to_encode.items())]
```
aaxelb, line 79:
```suggestion
        f'{key}="{value}"' for key, value in sorted(params_to_encode.items())
```
nit: don't need the intermediate list

addon_service/oauth1/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def generate_signature(
+    http_method: HTTPMethod,
+    url: str,
+    headers: dict,
+    oauth_consumer_secret: str,
+    oauth_token_secret: str | None = None,
+) -> str:
+    params_str = "&".join([f"{key}={headers[key]}" for key in sorted([*headers])])
```
aaxelb, line 107:
```suggestion
    params_str = "&".join(f"{key}={value}" for key, value in sorted(headers.items()))
```
nits: can skip the intermediate list and simplify as in `_construct_params` above

addon_service/oauth1/models.py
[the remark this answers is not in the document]
aaxelb: fwiw, even tho they overlap in concepts i think i'd prefer separate `OAuth1ClientConfig` and `OAuth2ClientConfig` for clarity (oauth terminology is low enough on clarity as it is)

addon_service/oauth1/models.py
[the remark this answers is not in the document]
aaxelb: at risk of hyper-bikeshedding... i'm gently in favor of `url` when it will be used to make requests and must resolve or something's wrong (as for `auth_url`) -- in contrast to `user_uri`/`resource_uri` that we use as identifiers but don't send requests to directly

aaxelb (changes requested):
looks good so far! had some suggestions -- most breakingly, adding a new model for the "temporary" oauth1 credentials

addon_service/authorized_storage_account/models.py
```diff
@@ -68,12 +72,21 @@ class AuthorizedStorageAccount(AddonsServiceBaseModel):
         blank=True,
         related_name="authorized_storage_account",
     )
+    _temporary_oauth1_credentials = models.OneToOneField(
+        "addon_service.ExternalCredentials",
```
aaxelb, line 76: instead of using `ExternalCredentials` for the temporary token/secret, how about move the encryption logic to a reusable abstract model and add a new table for oauth1 use?

```py
# addon_service/common/encrypted_json_model.py?
class EncryptedJsonModel(AddonsServiceBaseModel):
    class Meta:
        abstract = True

    encrypted_json = models.BinaryField()
    _salt = models.BinaryField()
    ... # encryption-related fields and methods from ExternalCredentials


# addon_service/credentials/models.py
class ExternalCredentials(EncryptedJsonModel):
    # credentials-specific methods only
    ...


# addon_service/oauth1/models.py
class Oauth1TemporaryToken(EncryptedJsonModel):
    # oauth1-specific methods/fields only
    ...
```

opaduchak: Can you elaborate on which particular logic you want to have encapsulated in these classes?

addon_service/oauth1/views.py
```diff
@@ -0,0 +1,42 @@
...
+    account.is_oauth1_ready = True
+    account.save()
+    update_account_with_additional_data(account, other_info)
+    return HttpResponse(status=HTTPStatus.OK)  # TODO: redirect
+
+
+def update_account_with_additional_data(account: AuthorizedStorageAccount, data: dict):
+    match account.external_service.int_addon_imp:
+        case AddonImpNumbers.ZOTERO_ORG:
+            account.external_account_id = data["userID"]
```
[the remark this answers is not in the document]
opaduchak: I agree, but in this particular case it would create more problems than solve, and this thing would have only one usage here, and getting external account id from this dict is specific only to Zotero.

addon_service/oauth1/models.py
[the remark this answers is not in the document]
jwalz: (recording result of previous discussion: we want to use uri/iri for things that are purely references, and url for things that will be used for requests, @aaxelb can own this as part of code cleanup tasks)

addon_service/authorized_storage_account/models.py
aaxelb: yep, elaborated the idea (tho not yet to working state) in branch `oauth1-elaboration` -- see `EncryptedDataclassModel`: https://github.com/opaduchak/gravyvalet/pull/1/files#diff-027727e35af18734580549b67475e3bd1d00ba714d205291c254e34eabf32abc

addon_service/oauth1/views.py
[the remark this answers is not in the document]
aaxelb: i don't think the enum addresses my concern; addon_service is still statically aware of specific imps...

this is what i'm thinking now, trying to keep imps contained without causing too much bother: https://github.com/opaduchak/gravyvalet/pull/3 (making `get_external_account_id` an instance method on `AddonImp`, so it can use `self.network` or `self.s3_client` if it needs, but also receives any "extra" key-value pairs from however auth succeeded (so zotero can find its `userID`))

addon_service/authorized_storage_account/models.py
aaxelb: i'm fine letting this idea go -- the current implementation does well enough keeping temporary credentials logically separate, even while sharing the ExternalCredentials table, and since [zotero does oauth1 (maybe) slightly wrong](https://github.com/opaduchak/gravyvalet/pull/1#discussion_r1674513336) we can't do the lookup-by-temporary-token that was part of the motivation for an oauth1-specific table

addon_toolkit/constrained_network/http.py
```diff
@@ -61,7 +61,7 @@ class HttpRequestor(typing.Protocol):
     def response_info_cls(self) -> type[HttpResponseInfo]: ...
 
     # abstract method for subclasses
-    def do_send(
+    def _do_send(
```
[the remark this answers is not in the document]
aaxelb: this is still unaddressed

(assuming it's a workaround for an error in `GravyvaletHttpRequestor.do_send`?) fix might be as simple as changing [this line](https://github.com/CenterForOpenScience/gravyvalet/blob/7db1040a51b94a1f8ee2f544532a189a29581165/addon_service/common/network.py#L89) to
```
if _private.account.credentials_format == CredentialsFormats.OAUTH2 and _response.status == HTTPStatus.UNAUTHORIZED:

File: addon_toolkit/credentials.py [hunk 23]
```diff
@@ -27,6 +27,36 @@ class AccessKeySecretKeyCredentials(Credentials):
...
+    oauth_token_secret: str
+
+    def iter_headers(self) -> typing.Iterator[tuple[str, str]]:
+        """
+        This is Zotero specific as other OAuth1.0a clients require request signing,
+        as per current architecture, we cannot it here.
+        """
+
+        yield "Authorization", f"Bearer {self.oauth_token_secret}"
+        # TODO: implement request signing for OAuth1.0a services that require it
```
aaxelb, line 42: gah, i missed that zotero uses the secret as bearer token instead of signing the request... feels wrong, but maybe it's true that `https` makes it safe, as argued in [their example client code](https://www.zotero.org/support/dev/web_api/v3/oauth#example_php):
```
/** OAuth support for all api requests may be added in the future
* but for now secure https provides similar benefits anyway
*/
```

(this speaks to the need for modular/swappable request-auth logic (instead of `iter_headers`) but sounds like that'll be part of s3 work -- this TODO seems fine for now)

addon_service/oauth1/models.py
```diff
@@ -0,0 +1,33 @@
...
+    request_token_url = models.URLField(null=False)
+    # URI to which user will be redirected to authenticate
+    auth_url = models.URLField(null=False)
+    # URI to which user will be redirected after authentication
+    auth_callback_url = models.URLField(null=False)
+    # URI to obtain access token
+    access_token_url = models.URLField(null=False)
+
+    client_key = models.CharField(null=True)
+    client_secret = models.CharField(null=True)
```
aaxelb, line 23: hmm, the client secret should be stored encrypted -- but since OAuth2ClientConfig is the same, maybe worth splitting that into another ticket? (...and i'm starting to reconsider the `EncryptedDataclassModel` abstract base, now that we have three potential uses for it...)

opaduchak: I agree, storing any unencrypted credentials in the db, poses security risks

addon_toolkit/credentials.py
opaduchak: We should be safe because `https` encypts entire `http` payload (including headers, query params, path, body, etc.)

addon_service/oauth1/models.py
jwalz: Yes, separate ticket. Need input from cloud eng/devops as to whether they want this in the database at all or if they'd prefer an approach like [this one](https://github.com/CenterForOpenScience/gravyvalet/pull/77)

PR #82, opened 2024-07-23
[ENG-5984] Extract shared Authorized[AddonType]Account code

opaduchak:
Extracted non storage-specific logic from AuthorizedStorageAccount to AuthorizedAccount
Other changes worth mentioning:
1. fixed `async with self.network.GET("/users/me") as _response:` throwing an exception because of path beginning with /
2. Some attrbuttes weren't extracted from Serializer and Model classes, because it would make code less maintainable


addon_service/abstract/authorized_account/models.py
```diff
@@ -0,0 +1,327 @@
...
+    external_account_id = models.CharField(null=False, blank=True, default="")
+    int_authorized_capabilities = models.IntegerField(
+        validators=[validate_addon_capability]
+    )
+    _api_base_url = models.URLField(blank=True)
+
+    account_owner = models.ForeignKey(
+        "addon_service.UserReference",
+        on_delete=models.CASCADE,
+        related_name="authorized_storage_accounts",
```
adlius, line 60: The `related_name` doesn't look right.

addon_service/abstract/authorized_account/models.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+    _credentials = models.OneToOneField(
+        "addon_service.ExternalCredentials",
+        on_delete=models.CASCADE,
+        primary_key=False,
+        null=True,
+        blank=True,
+        related_name="authorized_storage_account",
```
adlius, line 68: Same here.

addon_service/abstract/authorized_account/models.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+    _temporary_oauth1_credentials = models.OneToOneField(
+        "addon_service.ExternalCredentials",
+        on_delete=models.CASCADE,
+        primary_key=False,
+        null=True,
+        blank=True,
+        related_name="temporary_authorized_storage_account",
```
adlius, line 76: Same here.

addon_service/abstract/authorized_account/models.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+    oauth2_token_metadata = models.ForeignKey(
+        "addon_service.OAuth2TokenMetadata",
+        on_delete=models.CASCADE,  # probs not
+        null=True,
+        blank=True,
+        related_name="authorized_storage_accounts",
+        related_query_name="%(class)s_authorized_storage_account",
```
adlius, line 84: Same here.

adlius (changes requested):
Some minor nitpicks.

addon_service/addon_imp/instantiation.py
```diff
@@ -1,18 +1,23 @@
...
+    from addon_service.models import AuthorizedStorageAccount
+
+
 async def get_storage_addon_instance(
-    imp_cls: type[StorageAddonImp],
-    account: AuthorizedStorageAccount,
-    config: StorageConfig,
+    imp_cls: "type[StorageAddonImp]",
+    account: "AuthorizedStorageAccount",
+    config: "StorageConfig",
```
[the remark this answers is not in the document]
opaduchak: This is used in conjunction with conditional import (line 13) to avoid circular import, and get typing benefits

addon_imps/storage/box_dot_com.py
```diff
@@ -13,7 +13,7 @@ class BoxDotComStorageImp(storage.StorageAddonImp):
     """
 
     async def get_external_account_id(self, auth_result_extras: dict[str, str]) -> str:
-        async with self.network.GET("/users/me") as _response:
+        async with self.network.GET("users/me") as _response:
```
jwalz, line 16: Can you document the "why" of this change. Putting it in the CR description is fine.

addon_service/abstract/authorized_account/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+        validated_data: dict,
+    ) -> ExternalStorageService:  # TODO: change to ExternalService once implemented
+        ...
+
+    @abstractmethod
+    def create_authorized_account(
+        self,
+        external_service: ExternalStorageService,
+        validated_data: dict,
+    ) -> AuthorizedAccount: ...
```
jwalz, line 73:
```suggestion
    ) -> AuthorizedAccount:
    """Handles creation of the appropriate AuthorizedAccount subclass"""
```

addon_service/abstract/authorized_account/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+        "configured_storage_addons": "addon_service.serializers.ConfiguredStorageAddonSerializer",
+        "authorized_operations": "addon_service.serializers.AddonOperationSerializer",
+    }
+
+    @abstractmethod
+    def create_external_service(
+        self,
+        validated_data: dict,
+    ) -> ExternalStorageService:  # TODO: change to ExternalService once implemented
+        ...
```
jwalz, line 66:
```suggestion
        """Handles retrieval of the appropriate ExternalService instance from the validated serializer data."""
```

addon_service/abstract/authorized_account/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+class AuthorizedAccountSerializer(serializers.HyperlinkedModelSerializer):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Check whether subclasses declare all of required fields
+        if not REQUIRED_FIELDS.issubset(set(self.fields.keys())):
+            raise Exception(
+                f"{self.__class__.__name__} requires {self.REQUIRED_FIELDS} to be instantiated"
+            )
```
jwalz, line 34: honest question: These are already in the Meta's `fields`, won't that blow up instantiation for is if they're not present?

addon_service/addon_operation_invocation/models.py
```diff
@@ -84,10 +84,11 @@ def clean_fields(self, *args, **kwargs):
                 {"thru_addon": "thru_addon and thru_account must agree"}
             )
 
+    @property
```
jwalz, line 87: nit: if this is being turned into a property, please relocate it with the other properties

addon_service/authorized_storage_account/serializers.py
```diff
@@ -69,12 +38,18 @@ def __init__(self, *args, **kwargs):
...
         required=False,
     )
+    url = serializers.HyperlinkedIdentityField(
+        view_name=view_names.detail_view(RESOURCE_TYPE), required=False
+    )
+    account_owner = ReadOnlyResourceRelatedField(
+        many=False,
+        queryset=UserReference.objects.all(),
+        related_link_view_name=view_names.related_view(RESOURCE_TYPE),
+    )
```
jwalz, line 48: Out of curiosity, what prevented these from being included in the base serializer? Could you add that to the PR description so that we can all learn?

addon_service/migrations/0001_initial.py
```diff
@@ -381,6 +382,7 @@ class Migration(migrations.Migration):
                 null=True,
                 on_delete=django.db.models.deletion.CASCADE,
                 related_name="authorized_storage_accounts",
+                related_query_name="%(class)s_authorized_storage_account",
```
jwalz, line 385: need to re-make migrations?

jwalz (changes requested):

addon_service/abstract/authorized_account/serializers.py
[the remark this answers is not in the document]
opaduchak: This would be the case if there was a need to create `validated_data` dict, but in current case `validated_data` is an instance of validated data directly passed from serializer

addon_service/authorized_storage_account/serializers.py
opaduchak: `RESOURCE_TYPE` is dependent on dataclass which is dependent on subclass.
Surely, we can abstract this one, but it'd be a lot more messy, and will look as follows:
``` 
def AuthorizedAccountSerilzer(RESOURCE_TYPE: str):
    class _AuthroizedAccountSerilzer(serializers.HyperlinkedModelSerializer):
    ...
     account_owner = ReadOnlyResourceRelatedField(
        many=False,
        queryset=UserReference.objects.all(),
        related_link_view_name=view_names.related_view(RESOURCE_TYPE),
    )
```
and subclass it as follows

```
RESOURCE_TYPE = get_resource_type_from_model(AuthorizedStorageAccount)

class AuthroizedStorageAccountSerializer(AuthorizedAccountSerilzer(RESOURCE_TYPE)):

```

This is more DRY, drastically but more complex, obscure and hard to maintain.

addon_service/migrations/0001_initial.py
opaduchak: ORM says that no changes were detected

addon_service/abstract/authorized_account/serializers.py
opaduchak: No, it won't, tried this in interpreter. Fields in Meta are desingned to constrain inferring fields from model
image

addon_imps/storage/box_dot_com.py
opaduchak: The reason is that it raised exception. `ClientSession` expects path without / in the beginning

addon_service/abstract/authorized_account/serializers.py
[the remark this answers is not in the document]
jwalz: I guess what I'm saying is that the `create_authorized_account` logic doesn't/shouldn't need to be aware of the structure of `validated_data`.

So this would be
`def create_authorized_account(self, **kwargs): -> AuthorizedAccount...` and the base `create` function could call `create_authorized_account(**validated_data)`

Then the subclass could define
```
def create_authorized_account(
    self,
    external_storage_service, 
    display_name,
    authorized_capabilities,
    api_base_url,
    **kwargs,
): -> AuthorizedStorageAccount
```
The parameter names would need to align with the serializer field names (since those should align with the model field names), but that doesn't feel problematic to me. Would also potentially remove the need for `get_external_account`.

addon_service/abstract/authorized_account/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+                    authorized_account.pk
+                )
+            else:
+                raise serializers.ValidationError(
+                    {
+                        "initiate_oauth": "this external service is not configured for oauth"
+                    }
+                )
+        elif validated_data.get("credentials"):
+            authorized_account.credentials = validated_data["credentials"]
```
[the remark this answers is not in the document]
opaduchak: Fixed, hope this works

addon_service/authorized_storage_account/serializers.py
```diff
@@ -83,72 +59,35 @@ def __init__(self, *args, **kwargs):
...
-    def create(self, validated_data):
+    @staticmethod
+    def get_external_service(
+        external_storage_service: ExternalStorageService, **kwargs
+    ) -> ExternalStorageService:
+        return external_storage_service
+
+    def create_authorized_account(
+        self,
+        external_service: ExternalStorageService,
```
jwalz, line 70: I think that if we just make this `external_storage_service`, we can kill `get_external_service` as a whole.

jwalz (commented):
One last suggestion, looks good overall

addon_service/authorized_storage_account/serializers.py
opaduchak: We need external_service to set auth credentials in other methods, so I beg to disagree.

jwalz: `process_and_set_auth` has access to the `authorized_account`, which has access to its `external_service` (and to its `credentials_format`)

opaduchak: In this case I'll remove `process_and_set_auth` because `create` has become really thin

opaduchak: But in this case methods in base class should know what to get:
`external_storage_service`, `external_citations_service`, `external_remote_compute_service` which is unpreferable

jwalz: I'm confused.

The only access to `external_service` I'm seeing in `process_and_set_auth` is to determine the `credentials_format`.

Where would the base class need to know the specific model? Wouldn't the base class just use the `external_service` property that exists for exactly this purpose (or the `credentials_format` property that  similarly references the `external_service` property)?

addon_service/migrations/0001_initial.py
jwalz: Then what's the source of this change?

addon_service/authorized_storage_account/serializers.py
opaduchak: My bad, I have overlooked this

jwalz: No worries, that's what CR is for ;)

opaduchak: Fixed, take a look

addon_service/migrations/0001_initial.py
jwalz: (this is the one lingering thing I'm curious about before I merge)

jwalz (approved):
lgtm

PR #83, opened 2024-07-23
 [ENG-5985] Extract shared EXtrenal[AddonType]Service code 

opaduchak:
Extracted non storage-specific logic from ExternalStorageService to ExternalService
dependent on #82


addon_service/common/network.py
```diff
@@ -60,7 +61,7 @@ def __init__(
         *,
         client_session: aiohttp.ClientSession,
         prefix_url: str,
-        account: db.AuthorizedStorageAccount,
+        account: "db.AuthorizedStorageAccount",
```
adlius, line 64: Why stringified?

addon_service/common/network.py
```diff
@@ -144,7 +145,7 @@ class _PrivateNetworkInfo(_PrivateInfo):
 
     # keep network constraints away from imps
     prefix_url: str
-    account: db.AuthorizedStorageAccount
+    account: "db.AuthorizedStorageAccount"
```
adlius, line 148: Same here.

addon_service/external_storage_service/models.py
```diff
@@ -1,106 +1,23 @@
...
 
 
-class ExternalStorageService(AddonsServiceBaseModel):
-    display_name = models.CharField(null=False)
+class ExternalStorageService(ExternalService):
     int_addon_imp = models.IntegerField(
         null=False,
         validators=[validate_storage_imp_number],
         verbose_name="Addon implementation",
     )
```
adlius, line 12: Can this be moved to the `ExternalService` class as well? I assume each external service, regardless of type, should have a addon implementation and therefore a int representing that implementation.

addon_service/abstract/external_storage/models.py
```diff
@@ -0,0 +1,90 @@
...
+        validators=[validate_service_type],
+        verbose_name="Service type",
+    )
+    supported_scopes = ArrayField(models.CharField(), null=True, blank=True)
+    api_base_url = models.URLField(blank=True, default="")
+
+    oauth1_client_config = models.ForeignKey(
+        "addon_service.OAuth1ClientConfig",
+        on_delete=models.SET_NULL,
+        related_name="external_storage_services",
```
adlius, line 35: `related_name` seems wrong.

addon_service/abstract/external_storage/models.py
```diff
@@ -0,0 +1,90 @@
...
+        null=True,
+        blank=True,
+    )
+
+    oauth2_client_config = models.ForeignKey(
+        "addon_service.OAuth2ClientConfig",
+        on_delete=models.SET_NULL,
+        related_name="external_storage_services",
```
adlius, line 43: same here

adlius (changes requested):
Some small nitpicks.

addon_service/external_storage_service/models.py
opaduchak: This field has a validator which is checking whether this is **Storage** implementation, so it cannot be common for all external services.

adlius (approved):
Looks good.

PR #87, opened 2024-08-06
[ENG-6081] Unit tests failing in CI

opaduchak:
pinned attrs to 23.2.0, which raised deprecation warning and failed whole test suite

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -10,3 +10,4 @@ jsonschema==4.21.1
 aiohttp==3.9.3
 pyjwe==1.0.0
 cryptography==42.0.7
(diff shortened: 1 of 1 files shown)


adlius (approved):

jwalz:
Why/where are we using `attrs`? Is this coming from one of our other dependencies, and should we be updating that dependency if so? Would be nice to switch to Poetry to have a better view into this >_<

opaduchak:
It seems that `attrs` fixed this issue on their side in version [24.2.0](https://github.com/python-attrs/attrs/releases/tag/24.2.0), so this PR is no longer needed. Therefore, I am closing this PR

PR #96, opened 2024-08-08
Citations infrastructure

opaduchak:
Implement `AuthorizedCitationAccount`, `ExternalCitationService`, `ConfiguredCitationAddon`, and glue them together for them to work properly.


addon_service/authorized_citation_account/models.py
```diff
@@ -0,0 +1,81 @@
+from django.db import models
+
+from addon_service.abstract.authorized_account.models import AuthorizedAccount
+from addon_service.addon_imp.instantiation import get_citation_addon_instance
+from addon_service.external_citation_service.models import ExternalCitationService
+from addon_toolkit.interfaces.citation import CitationConfig
+
+
+class AuthorizedCitationAccount(AuthorizedAccount):
+    """Model for describing a user's account on an ExternalService.
```
adlius, line 10:
```suggestion
    """Model for describing a user's account on an ExternalCitationService.
```

addon_service/authorized_citation_account/models.py
```diff
@@ -0,0 +1,81 @@
...
+            self,
+            self.citation_imp_config,
+        )
+        self.external_account_id = await imp.get_external_account_id(auth_extras or {})
+        await self.asave()
+
+    @property
+    def citation_imp_config(self) -> CitationConfig:
+        return CitationConfig(
+            max_upload_mb=self.external_service.max_upload_mb,
```
adlius, line 77: I have a feeling that the `max_upload_mb` may not be relevant for a citation addon, since it is not meant to be a storage service.

addon_service/authorized_citation_account/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            return AuthorizedCitationAccount.objects.create(
+                _display_name=display_name,
+                external_citation_service=external_citation_service,
+                account_owner=account_owner,
+                authorized_capabilities=authorized_capabilities,
+                api_base_url=api_base_url,
+            )
+        except ModelValidationError as e:
+            raise serializers.ValidationError(e)
```
adlius, line 83: Seeing that this is mostly copy-pasted from `AuthorizedStorageAccountSerializer`, can we somehow DRY it up a little bit? Maybe put it in `AuthorizedAccountSerializer`? Not a huge concern though.

addon_service/configured_citation_addon/serializers.py
```diff
@@ -0,0 +1,62 @@
...
+from addon_service.configured_citation_addon.models import ConfiguredCitationAddon
+
+
+RESOURCE_TYPE = get_resource_type_from_model(ConfiguredCitationAddon)
+
+
+class ConfiguredCitationAddonSerializer(ConfiguredAddonSerializer):
+    """api serializer for the `ConfiguredCitationAddon` model"""
+
+    root_folder = serializers.CharField(required=False, allow_blank=True)
```
adlius, line 21: I don't think `root_folder` is needed on the `ConfiguredCitationAddonSerializer`. It shouldn't be needed on the `ConfiguredStorageAddonSerializer` as well.

addon_service/external_citation_service/models.py
```diff
@@ -0,0 +1,35 @@
+from django.db import models
+
+from addon_service.abstract.external_storage.models import ExternalService
+
+
+class ExternalCitationService(ExternalService):
+    int_addon_imp = models.IntegerField(
+        null=False,
+        verbose_name="Addon implementation",
```
adlius, line 9: Do we need a validator here?

addon_service/external_citation_service/models.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    # Do we need the wb_key field for ExternalCitationService ?
+    wb_key = models.CharField(null=False, blank=True, default="")
```
adlius, line 12: `wb_key` should not be necessary here since there are no file ops.

adlius (changes requested):
Some questions and nitpicks. Overall looks great!

addon_service/configured_citation_addon/serializers.py
opaduchak: we need `root_folder`, as we have to bind our node to certain collection. Or do you imply that this should be declared somewhere else?

addon_service/external_citation_service/models.py
opaduchak: @jwalz stated that we need `wb_key` inside `ExternalCitationService`, for us to map credentials from `osf` addons to `GV` ones. We don't need wb_key inside `CitationConfig` class. `wb_key` is quite misleading

addon_service/authorized_citation_account/serializers.py
opaduchak: We require different arguments for different subclasses of `AuthorizedAccount`, so DRY implementation wouldn't be clear this way

addon_service/configured_citation_addon/serializers.py
adlius: Yes, we need to bind our node to certain collection. But `root_folder` defined here would imply that all the nodes that are tied to this citation addon share the same `root_folder`. I don't think this is correct. The root folder should be different for each node and that's why it is already on `AuthorizedStorageAccount` model.

opaduchak: Ok, I've dug even deeper, and it seems that `ConfiguredCitationAddon` has 1 to 1 relation with Node, so for each Node new `ConfiguredCitationAddon` has to be created. In case with AuthorizedAccount, you are 100% right.

PR #102, opened 2024-08-20
[ENG-5973] Support serializing and modifying "supported_features" for ExternalStorageService

adlius:


addon_service/external_storage_service/models.py
```diff
@@ -53,16 +52,19 @@ class ExternalStorageService(ExternalService):
...
     )
 
     @property
     def supported_features(self) -> list[SupportedFeatures]:
         """get the enum representation of int_supported_features"""
-        return [SupportedFeatures(item) for item in self.int_supported_features]
+        return SupportedFeatures(self.int_supported_features)
+
+    @supported_features.setter
+    def authorized_capabilities(self, new_supported_features: SupportedFeatures):
```
jwalz, line 65: oops
```suggestion
    def supported_features(self, new_supported_features: SupportedFeatures):
```

adlius: Oops. But I don't think this is needed at all right now because we only modify/set the supported features field through the admin app, and the admin app directly modifies the `int_supported_features` field on the model.

PR #107, opened 2024-08-29
[ENG-6041] Added Google Drive implementation

opaduchak:


addon_toolkit/interfaces/storage.py
```diff
@@ -37,7 +37,7 @@ class StorageConfig:
     external_account_id: str | None = None
 
 
-class ItemType(enum.Enum):
+class ItemType(enum.StrEnum):
```
adlius, line 40: What's the purpose of turning this into StrEnum?

addon_toolkit/interfaces/citation.py
```diff
@@ -31,7 +31,7 @@ class CitationConfig:
     external_account_id: str | None = None
 
 
-class ItemType(Enum):
+class ItemType(StrEnum):
```
adlius, line 34: Same here

adlius (commented):
Some questions

addon_imps/storage/google_drive.py
```diff
@@ -0,0 +1,100 @@
...
+)
+
+
+class GoogleDriveStorageImp(storage.StorageAddonHttpRequestorImp):
+    """storage on google drive
+
+    see https://developers.google.com/drive/api/reference/rest/v3/
+    """
+
+    async def get_external_account_id(self, _: dict[str, str]) -> str:
```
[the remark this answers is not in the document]
opaduchak: I implemented this method because it is being called outside implementation. Othewise another parts of application will fail

addon_toolkit/interfaces/storage.py
opaduchak: This fixes filters for `get_root_items` and `list_child_items`

addon_toolkit/interfaces/citation.py
opaduchak: This fixes filters for `get_root_items` and `list_child_items`

addon_toolkit/iri_utils.py
```diff
@@ -89,23 +90,28 @@ def as_query_string(self) -> str:
 
 
 def iri_with_query(iri: str, query_params: KeyValuePairs | Multidict) -> str:
```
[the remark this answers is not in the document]
opaduchak: So, GoogleDrive requires additional query parameters during OAuth2.0 flow to retrieve `refresh_token`. In previous behaviour, after these query params were embedded into `auth_url` during configuration on admin page, they were removed here, which caused flow to fail.

This change doesn't break OAuth2.0 flow for other providers, and in my opinion this behaviour is the most desirable and expectable.

I may have fixed this putting new field in OAuth2ClientConfig, but I found previous approach to be the simplest and easiest one.

addon_service/abstract/configured_addon/models.py
```diff
@@ -49,7 +49,8 @@ def display_name(self):
 
     @display_name.setter
     def display_name(self, value: str):
-        self._display_name = value
+        if value:
+            self._display_name = value
```
ly-mariia, line 54: maybe use this approach will be better + typing fix

```
def display_name(self, value: str) -> str | None:
    return  self._display_name if value else None
```

ly-mariia (changes requested):

addon_service/abstract/configured_addon/models.py
opaduchak: this is a setter, therefore it returns nothing

ly-mariia:
LGTM

ly-mariia (approved):

addon_service/addon_imp/instantiation.py
```diff
@@ -33,15 +34,19 @@ async def get_storage_addon_instance(
...
-        network=GravyvaletHttpRequestor(
-            client_session=await get_singleton_client_session(),
-            prefix_url=config.external_api_url,
-            account=account,
-        ),
-    )
+    assert issubclass(imp_cls, StorageAddonImp)
+    assert (
+        imp_cls is not StorageAddonImp
+    ), "Addons shouldn't directly extent storage addon imp, but "
```
brianjgeiger, line 40: Nit:
```suggestion
    ), "Addons shouldn't directly extend storage addon imp, but "
```

addon_toolkit/imp.py
```diff
@@ -87,7 +87,9 @@ async def invoke_operation(
         if not inspect.iscoroutinefunction(_operation_method):
             _operation_method = sync_to_async(_operation_method)
         _result = await _operation_method(**_kwargs)
-        assert isinstance(_result, operation.result_dataclass)
+        assert isinstance(
+            _result, operation.result_dataclass
+        ), f"expected {operation.result_dataclass.__name__} type to bre returned from method {_operation_method.__name__}, got {_result.__class__.__name__}"
```
brianjgeiger, line 92: Nit:
```suggestion
        ), f"expected {operation.result_dataclass.__name__} type to be returned from method {_operation_method.__name__}, got {_result.__class__.__name__}"
```

brianjgeiger (approved):
Two minor typos, otherwise looks good to me.

addon_service/common/known_imps.py
```diff
@@ -73,6 +75,7 @@ class AddonImpNumbers(enum.Enum):
     BOX_DOT_COM = 1001
     ZOTERO_ORG = 1002
     S3 = 1003
+    GOOGLE_DRIVE = 1004
```
brianjgeiger, line 78: From the CR of Figshare, I gather this is going to conflict with Mendeley.

PR #110, opened 2024-09-10
[ENG-6038] Implement FigshareStorageAddonImp

opaduchak:
Implemented FigshareAddonImp

### DIFF
--- a/addon_imps/storage/figshare.py
+++ b/addon_imps/storage/figshare.py
@@ -1,0 +1,223 @@
+from __future__ import annotations
+
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from dataclasses import (
+    dataclass,
+    fields,
+)
+
+from twisted.internet.protocol import Protocol
+
+from addon_toolkit.async_utils import join
+from addon_toolkit.interfaces import storage
+from addon_toolkit.interfaces.storage import (
+    ItemResult,
+    ItemSampleResult,
+    ItemType,
+)
+
+
+FILE_CONTAINING_ARTICLES = (3, 4)
+
+
+class FigshareStorageImp(storage.StorageAddonHttpRequestorImp):
+    """storage on figshare
+
+    see https://developers.google.com/drive/api/reference/rest/v3/
+    """
+
+    async def get_external_account_id(self, _: dict[str, str]) -> str:
+        return ""
+
+    async def list_root_items(self, page_cursor: str = "") -> storage.ItemSampleResult:
+        page_cursor = int(page_cursor or 1)
+        items = await join(
+            self._fetch_projects(page_cursor),
+            self._fetch_articles(page_cursor),
+        )
+
+        return ItemSampleResult(
+            items=[entry.item_result for entry in items],
+            next_sample_cursor=str(page_cursor + 1),
+        )
+
+    async def get_item_info(self, item_id: str) -> storage.ItemResult:
+        if not item_id:
+            return ItemResult(item_id="", item_name="", item_type=ItemType.FOLDER)
+        if self._is_article_id(item_id):
+            result = await self._fetch_article(item_id)
+        elif self._is_project_id(item_id):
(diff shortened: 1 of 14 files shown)


addon_imps/storage/figshare.py
```diff
@@ -0,0 +1,216 @@
...
+        assert project_id.startswith("projects/")
+
+        async with self.network.GET(
+            f"account/{project_id}",
+        ) as response:
+            return Project.from_json(await response.json_content())
+
+
+###
+# module-local helpers
```
sh-andriy, line 154: I wonder if this comment necessary, noticed that you have the same ones later on?

addon_service/common/known_imps.py
```diff
@@ -73,6 +77,8 @@ class AddonImpNumbers(enum.Enum):
     BOX_DOT_COM = 1001
     ZOTERO_ORG = 1002
     S3 = 1003
+    GOOGLE_DRIVE = 1004
```
sh-andriy, line 80: I know we have this number for **Mendeley**, I think you can just change this to `1005`, or/and Figshare to `1007`
`MENDELEY = 1004`

addon_service/common/network.py
```diff
@@ -92,6 +92,8 @@ async def _try_send(self, request: HttpRequestInfo):
             request.http_method,
             _url,
             headers=await _private.get_headers(),
+            params=request.query,
+            json=request.json,
             # TODO: content
```
sh-andriy, line 97: Are these **TODO's** you currently have doable or if not do we need to include all of them if we won't be doing them?

addon_toolkit/interfaces/citation.py
```diff
@@ -31,7 +31,7 @@ class CitationConfig:
     external_account_id: str | None = None
 
 
-class ItemType(Enum):
+class ItemType(StrEnum):
```
sh-andriy, line 34: I'm curious will this change affect the `Zotero` and `Mendeley API` implementations? If not leave it as you did

sh-andriy (changes requested):
Overall looks pretty good, just need to clarify some points that I commented

addon_toolkit/interfaces/citation.py
opaduchak: This has been done to fix filters invokation from frontend

addon_service/common/network.py
opaduchak: These TODO's had been here before me


I'd leave them as there might be need to include urlencoded data or something else.

sh-andriy: Oh okay, if they had been here before than leave it as it is

addon_service/common/known_imps.py
opaduchak: This is Google drive related

addon_imps/storage/figshare.py
opaduchak: The same comment was in different file

sh-andriy (approved):
LGTM

addon_toolkit/imp.py
```diff
@@ -87,7 +87,9 @@ async def invoke_operation(
         if not inspect.iscoroutinefunction(_operation_method):
             _operation_method = sync_to_async(_operation_method)
         _result = await _operation_method(**_kwargs)
-        assert isinstance(_result, operation.result_dataclass)
+        assert isinstance(
+            _result, operation.result_dataclass
+        ), f"expected {operation.result_dataclass.__name__} type to bre returned from method {_operation_method.__name__}, got {_result.__class__.__name__}"
```
brianjgeiger, line 92: This was mentioned in the Google Drive review, but 
```suggestion
        ), f"expected {operation.result_dataclass.__name__} type to be returned from method {_operation_method.__name__}, got {_result.__class__.__name__}"
```

brianjgeiger (approved):
Seems good to me

PR #111, opened 2024-09-12
[ENG-6037] Dropbox implementation 

Vlad0n20:


addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,113 @@
...
+                    "path": "/",
+                    "recursive": True,
+                },
+            ) as _response:
+                _parsed = _DropboxParsedJson(await _response.json_content())
+                return storage.ItemSampleResult(
+                    items=list(_parsed.item_results()),
+                    total_count=len(_parsed.response_json["entries"]),
+                    next_sample_cursor=_parsed.response_json["cursor"],
+                )
```
opaduchak, line 43: so, if we pass not falsy `item_id` , it will return nothing?

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,113 @@
...
+    ) -> bool:
+        return self.ITEM_TYPE[item_json["type"]] == item_type
+
+    def item_results(
+        self,
+        item_type: storage.ItemType | None = None,
+    ) -> typing.Iterator[storage.ItemResult]:
+        for _item in self.response_json["entries"]:
+            if (item_type is None) or self._item_has_type(_item, item_type):
+                yield self._parse_item(_item)
```
opaduchak, line 113: Is there no way to filter it on DropBox side?
Just question

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,113 @@
...
+        return await self.get_item_info("")
+
+    async def get_item_info(self, item_id: str) -> storage.ItemSampleResult:
+        if not item_id:
+            async with self.network.POST(
+                "files/list_folder",
+                {
+                    "include_non_downloadable_files": True,
+                    "path": "/",
+                    "recursive": True,
```
opaduchak, line 35: what does recursive parameter mean? Would that return contents of subfolders? I don't believe this is the behaviour we'd want

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,113 @@
...
+    async def list_root_items(self, page_cursor: str = "") -> storage.ItemSampleResult:
+        if page_cursor:
+            async with self.network.POST(
+                "/files/list_folder/continue",
+                {"cursor": page_cursor},
+            ) as _response:
+                _parsed = _DropboxParsedJson(await _response.json_content())
+                return storage.ItemSampleResult(
+                    items=list(_parsed.item_results()),
+                )
```
opaduchak, line 24: don't we need to return `next_sample_cursor` or `total_count` here

opaduchak (changes requested):

addon_imps/storage/dropbox.py
Vlad0n20: I use this function only in one place, where I pass the item_id as '', so that the item_id will not falsy

addon_imps/storage/dropbox.py
Vlad0n20: I couldn't find a way to do this on  DropBox side

addon_imps/storage/dropbox.py
Vlad0n20: Yes, the recursive parameter means that returns the contents of subfolder. I thought that we need it, I'll remove that behaviour.

addon_imps/storage/dropbox.py
Vlad0n20: Oh yeh, I missed it

addon_imps/storage/dropbox.py
opaduchak: This is being used by frontend to fetch item details, in this case there will be non-falsy values
Moreover implementation mustn't assume how it's public API will be used

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
```
opaduchak, line 8: please remove all .idea files

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,118 @@
...
+        self,
+        item_id: str,
+        page_cursor: str = "",
+        item_type: storage.ItemType | None = None,
+    ) -> storage.ItemSampleResult:
+        if page_cursor:
+            async with self.network.POST(
+                "files/list_folder/continue",
+                json={"cursor": page_cursor},
+            ) as _response:
```
opaduchak, line 60: have you tested this case?

addon_service/migrations/0003_configuredcitationaddon_root_folder_and_more.py
```diff
@@ -1,4 +1,5 @@
-# Generated by Django 4.2.7 on 2024-09-10 13:51
+
+# Generated by Django 4.2.7 on 2024-09-10 13:10
```
opaduchak, line 2: please remove these unneeded diffs

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,118 @@
...
+                _parsed = _DropboxParsedJson(await _response.json_content())
+                return storage.ItemSampleResult(
+                    items=list(_parsed.item_results()),
+                    total_count=len(_parsed.response_json["entries"]),
+                    next_sample_cursor=_parsed.response_json["cursor"],
+                )
```
opaduchak, line 27: this may be replaced by list_child_items

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,118 @@
...
+    async def list_child_items(
+        self,
+        item_id: str,
+        page_cursor: str = "",
+        item_type: storage.ItemType | None = None,
+    ) -> storage.ItemSampleResult:
+        if page_cursor:
+            async with self.network.POST(
+                "files/list_folder/continue",
+                json={"cursor": page_cursor},
```
opaduchak, line 59: don't we still need path if we have cursor?

opaduchak (changes requested):

addon_imps/storage/dropbox.py
Vlad0n20: Cursor already have information about  needed path

addon_imps/storage/dropbox.py
Vlad0n20: done

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    async def get_external_account_id(self, _: dict[str, str]) -> str:
+        return ""
+
+    async def list_root_items(self, page_cursor: str = "") -> storage.ItemSampleResult:
+        if page_cursor:
+            return await self.list_child_items(item_id="", page_cursor=page_cursor)
+        result = await self.list_child_items("")
+        return result
```
opaduchak, line 20:
```suggestion
        return self.list_child_items(item_id="", page_cursor=page_cursor)
```

opaduchak, line 20: this if is unneeded, please just call `self.list_child_items`

opaduchak (changes requested):

addon_imps/storage/dropbox.py
```diff
@@ -0,0 +1,100 @@
...
+from addon_toolkit.interfaces.storage import ItemType
+
+
+class DropboxStorageImp(storage.StorageAddonHttpRequestorImp):
+    """storage on dropbox.com
+    see https://www.dropbox.com/developers/documentation/http/documentation
+    """
+
+    async def get_external_account_id(self, _: dict[str, str]) -> str:
+        return ""
```
adlius, line 14: Do we not need the external account id? Or does Dropbox not have external account id available?

adlius (approved):
LGTM. One question.

PR #126, opened 2024-10-11
[ENG-6042] [ENG-6259] - Implement OnedriveStorageAddonImp & Tests

ly-mariia:
Added new implementation for storage addon - OneDrive and tests for it

https://openscience.atlassian.net/browse/ENG-6042
https://openscience.atlassian.net/browse/ENG-6259


sh-andriy (approved):
Overall looks good!

addon_service/admin/__init__.py
```diff
@@ -23,9 +23,6 @@ class ExternalStorageServiceAdmin(GravyvaletModelAdmin):
         "int_addon_imp": known_imps.AddonImpNumbers,
         "int_credentials_format": CredentialsFormats,
         "int_service_type": ServiceTypes,
-    }
-
-    enum_multiple_choice_fields = {
```
brianjgeiger, line 28: Should we be getting rid of this, or does it behave differently in admin vs choice fields? It looks like it gives different controls compared to the enum_choice_fields. https://github.com/CenterForOpenScience/gravyvalet/blob/c197c7e6a81eb340e480733299d912b82dd07880/addon_service/admin/_base.py#L69-L79

brianjgeiger (changes requested):
One thing I think probably shouldn't be changed.

addon_service/admin/__init__.py
sh-andriy: good point, I will remove these changes

sh-andriy:
@brianjgeiger Updated from `develop` and resolved **failed checks**, tested It out now everything seems to be good

brianjgeiger (approved):

PR #129, opened 2024-10-16
[ENG-6353]Try to fix gv developer docs

adlius:


Dockerfile
```diff
@@ -55,8 +55,6 @@ FROM gv-base AS gv-deploy-deps
 RUN curl -sSL https://install.python-poetry.org | python3 - --version 1.8.3
 RUN python -m venv .venv
 RUN poetry install --without dev
-# copy auto-generated static docs (without the dev dependencies that built them)
-COPY --from=gv-docs /code/addon_service/static/gravyvalet_code_docs/ /code/addon_service/static/gravyvalet_code_docs/
 # collect static files into a single directory:
```
brianjgeiger, line 58: We could remove this line as well. It looks like it's not doing anything right now.

brianjgeiger (changes requested):
The tiniest of comments, otherwise looks like what we're expecting.

brianjgeiger (approved):

PR #131, opened 2024-10-16
[ENG-6039] GitHubImp

Vlad0n20:


addon_service/oauth2/utils.py
```diff
@@ -113,8 +114,15 @@ async def get_refreshed_access_token(
...
 ) -> FreshTokenResult:
+    import urllib.parse
+
     _client = await get_singleton_client_session()
     async with _client.post(token_endpoint_url, data=request_body) as _token_response:
+        response_text = await _token_response.text()
+        if "github.com" in token_endpoint_url:
+            response_data = dict(urllib.parse.parse_qsl(response_text))
+            response_data["expires_in"] = int(response_data["expires_in"])
+            return FreshTokenResult.from_token_response_json(response_data)
```
adlius, line 125: So here we have to do some special treatment for Github? Is there any way we can make this more generic?

adlius (commented):
One question.

addon_service/oauth2/utils.py
Vlad0n20: I had :
ContentTypeError at /v1/oauth2/callback/
0, message='Attempt to decode JSON with unexpected mimetype: application/x-www-form-urlencoded; charset=utf-8', url=URL('https://github.com/login/oauth/access_token')
Only for GitHub addon. Others work properly, maybe in the future appear new addons with problem like that but now only GitHub. I can rewrite it using try/exept for ContentTypeError it will be more generic.

opaduchak: I'd suggest doing it based on content-type, rather than using try-catch

adlius: @Vlad0n20 I agree with @opaduchak . Maybe we can try to do if statements on the content type and no having to special-case for GitHub.

adlius (approved):

PR #137, opened 2024-10-25
[ENG-6040] Gitlab imp

opaduchak:


addon_imps/storage/gitlab.py
```diff
@@ -0,0 +1,182 @@
...
+class GitlabStorageImp(storage.StorageAddonHttpRequestorImp):
+    """storage on gitlab
+
+    see https://developers.google.com/drive/api/reference/rest/v3/
+    """
+
+    async def get_external_account_id(self, _: dict[str, str]) -> str:
+        async with self.network.GET("user/preferences") as response:
+            resp_json = await response.json_content()
+            return resp_json.get("user_id", "")
```
sh-andriy, line 30: I'm not sure but isn't the **Gitlab** api returns `users id` using `self.network.GET("user")` and than just get `"id"` from response instead of `user/preferences` ? 
So it would look something like this:
```
    async with self.network.GET("user") as response:
        resp_json = await response.json_content()
        return str(resp_json["id"])
```

addon_imps/storage/gitlab.py
```diff
@@ -0,0 +1,182 @@
...
+    id: str
+    name: str
+
+    @property
+    def item_result(self) -> ItemResult:
+        return ItemResult(
+            item_id=f"{self.id}:",
+            item_name=self.name,
+            item_type=ItemType.FOLDER,
+        )
```
sh-andriy, line 182: I think you should add here `def from_json()` method to parse **`JSON data`** so It prevents`AttributeError` if there is not such method already, because you are calling this method here `items=[Repository.from_json(item).item_result for item in resp]`. If there is such method just leave It as It is

addon_imps/storage/gitlab.py
```diff
@@ -0,0 +1,182 @@
...
+            return Repository.from_json(content).item_result
+
+    async def get_file_or_folder(self, parsed_id: ItemId):
+        async with self.network.GET(f"projects/{parsed_id.repo_id}") as response:
+            content = await response.json_content()
+            ref = content.get("default_branch")
+        if file_item := await self._get_file(parsed_id, ref):
+            return file_item
+        # try to list files under folder, if it succeeds, proceed to return folder, else propagate the error
+        await self.list_child_items(parsed_id.raw_id)
```
sh-andriy, line 89: I think you can remove this comment

sh-andriy:
Overall looks good, just some points I commented about

sh-andriy (changes requested):

addon_imps/storage/gitlab.py
opaduchak: it extends ItemResultable, which inspects dataclass class and makes this automatically

addon_imps/storage/gitlab.py
opaduchak: This endpoint is `user/preferences`
I used this enpoint because it returns smaller json)
https://docs.gitlab.com/ee/api/users.html#get-your-user-preferences

addon_imps/storage/gitlab.py
opaduchak: I'd like to leave it to clarify why this call is being made

addon_imps/storage/gitlab.py
sh-andriy: Gotcha, seems reasonable) Leave it as it is then

sh-andriy (approved):

brianjgeiger (approved):

PR #139, opened 2024-10-30
[ENG-6437] GV: Add new flags to ItemResult class

opaduchak:
added `can_be_root` and `may_contain_root_candidates` properties for storage addons ItemResult


sh-andriy (approved):
Overall looks good

PR #142, opened 2024-10-30
fixed s3 implementation to return ids as full path

opaduchak:


brianjgeiger:
@opaduchak Tests broke again with your last commit.

PR #143, opened 2024-11-01
Implementd wb settings generation on GV side

opaduchak:


brianjgeiger (changes requested):
This looks good with the exception of the test failures.

brianjgeiger (approved):

PR #149, opened 2024-11-13
[ENG-5975] NewRelic Instrumentation

mfraezz:
# Purpose
Integrate NewRelic with Gravy.

# Changes
- Add `newrelic` dependency, update `poetry.lock`
- Initialize NewRelic agent where Sentry is also being initialized

# Notes
- There may be a better place for application-init calls to go.
- This will require Helm chart and config changes to take effect, but should be relatively low-risk.


brianjgeiger (approved):

mfraezz:
Merging this ahead of chart changes, so we'll be able to configure it when those are ready.

PR #159, opened 2024-11-20
[ENG-6551] Reset credentials during oauth flow

adlius:


addon_service/authorized_account/models.py
```diff
@@ -255,6 +255,8 @@ def initiate_oauth2_flow(self, authorized_scopes=None):
             ),
             state_nonce=oauth2_utils.generate_state_nonce(),
         )
+        # reset credentials
+        self._credentials = None
```
adlius, line 259: When we try to reconnect an addon, the `update` hook of the `AuthorizedAccountSerializer` would try to `initiate_oauth2_flow` again, but the problem is the credentials already exist. Therefore we should clear the credentials.

opaduchak (approved):
LGTM!

PR #168, opened 2024-11-28
[ENG-6628] Fixed github oauth app behaviour

opaduchak:


sh-andriy (approved):

addon_service/oauth2/models.py
```diff
@@ -94,11 +107,16 @@ def validate_nonce_and_token(self):
             raise ValidationError(
                 "Error on OAuth2 Flow: state nonce and refresh token both present."
             )
-        if not (self.state_nonce or self.refresh_token):
+        if not (self.state_nonce or self.access_token_only or self.refresh_token):
             raise ValidationError(
                 "Error in OAuth2 Flow: Neither state nonce nor refresh token present."
```
brianjgeiger, line 112: Similarly, I'm not sure that this message is quite right either.

brianjgeiger (changes requested):
Just a couple of error messages that might need updating with the logic changes, otherwise looks great.

addon_service/oauth2/models.py
opaduchak: These errors are valid if we don't have said quirk

brianjgeiger (approved):

PR #197, opened 2024-12-29
[ENG-6774] export terms and conditions for boa/computing addons

felliott:
# Changes

* Update `supported features` list for computing addons
* Make sure those supported features are properly serialized so the terms-and-conditions page can be rendered
* Fix up Computing addon instantiation code to correctly instantiate.


opaduchak (approved):
LGTM

PR #201, opened 2025-01-06
[ENG-6820] Fix n+1 queries

opaduchak:


addon_service/authentication.py
```diff
@@ -9,6 +9,8 @@ class GVCombinedAuthentication(drf_authentication.BaseAuthentication):
     """Authentication supporting session, basic, and token methods."""
 
     def authenticate(self, request: DrfRequest):
+        if request.session.get("user_reference_uri"):
```
adlius, line 12: I think this is a potential security problem. This session lives on gravyvalet. A `sessionId` cookie is set on the client side by whatever origin that hosts gravyvalet. That means when user logs out of osf.io, the `sessionId` cookie is not cleared on the client, nor is the session itself invalidated on the gravyvalet server side. Here's one potential scenario that could happen on staging:

1. User logs into staging.osf.io.
2. User makes request to gravyvalet.
3. gravyvalet sets the `sessionId` cookie.
4. User logs out of staging.osf.io.
5. `sessionId` cookie is not cleared on the client side, nor is the session invalidated on gravyvalet server side.
6. Using the same browser, an unauthenticated user successfully makes another request to gravyvalet using the old `sessionId` cookie.

Unless we can make sure that logging out of osf.io would also invalidated the corresponding session on the GV side, I don't think we can rely on just `user_reference_uri` on the session. Instead, we have to make sure the user provides the correct osf cookies for every request.

@opaduchak @brianjgeiger @aaxelb for a sanity check.

adlius (changes requested):
DB query optimizations looks good. I have one concern regarding sessions and cookies.

addon_service/authentication.py
brianjgeiger: Agreed, with no way to invalidate the session, we need to rely on osf.io for maintaining that information.

opaduchak: Would you like me to revert this change or implement session invalidation during logout?

adlius: Let's revert the change for this PR. If we want to implement session invalidation we can create a new ticket.

adlius (approved):

PR #202, opened 2025-01-08
[ENG-6007] Add apiBaseUrlOptions field

adlius:


brianjgeiger (approved):
Does this need anything to make it work in the admin app, or does it just happen? I presume that's where we are setting the urls.

addon_service/migrations/0010_externalservice_api_base_url_options_and_more.py
```diff
@@ -0,0 +1,36 @@
...
+        migrations.AlterField(
+            model_name="externalservice",
+            name="icon_name",
+            field=models.FilePathField(
+                blank=True,
+                match=".*\\.(jpg|png|svg)$",
+                null=True,
+                path=pathlib.PurePosixPath("/code/addon_service/static/provider_icons"),
+            ),
+        ),
```
opaduchak, line 35: Not related to this PR: it seems that base path for provider icons is being evaluated during migration time and not during runtime. We should either mitigate this or generate migrations only inside of docker-compose (and document this)

opaduchak (approved):
LGTM

adlius:
> Does this need anything to make it work in the admin app, or does it just happen? I presume that's where we are setting the urls.

There is a input field in the admin app.

PR #220, opened 2025-01-29
[ENG-7084] fixed dropbox migrate issue

opaduchak:


addon_service/common/base_model.py
```diff
@@ -18,7 +19,8 @@ def save(self, *args, **kwargs):
         if self._state.adding:
             self.created = timezone.now()
         self.modified = timezone.now()
-        self.full_clean()
+        if settings.AUTO_FULL_CLEAN:
+            self.full_clean()
```
adlius, line 23: Would it be better to pass a `full_clean` parameter to the `save()` hook instead of using a variable in settings?

adlius (commented):
One question.

adlius (approved):

PR #221, opened 2025-01-30
[ENG-7083] Add celery-beat; add periodic `refresh_addon_tokens` task

adlius:
The management command itself is ported from [osf.io](https://github.com/CenterForOpenScience/osf.io/blob/develop/scripts/refresh_addon_tokens.py)


app/settings.py
```diff
@@ -242,3 +245,20 @@
 AMQP_BROKER_URL = env.AMQP_BROKER_URL
 OSF_BACKCHANNEL_QUEUE_NAME = env.OSF_BACKCHANNEL_QUEUE_NAME
 GV_QUEUE_NAME_PREFIX = env.GV_QUEUE_NAME_PREFIX
+
+# Celery Beat
+CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"
+CELERY_BEAT_SCHEDULE = {
+    "refresh_addon_tokens": {
+        "task": "addon_service.management.commands.refresh_addon_tokens.refresh_addon_tokens",
+        "schedule": crontab(minute="*"),
```
opaduchak, line 254: Maybe we should settle for something like once a week, and override it locally if we need to test?

addon_service/management/commands/refresh_addon_tokens.py
```diff
@@ -0,0 +1,71 @@
...
+                account.oauth2_token_metadata.access_token_expiration.strftime("%c"),
+            )
+        )
+        if not fake:
+            if allowance < 1:
+                try:
+                    time.sleep(rate_limit[1] - (time.time() - last_call))
+                except (ValueError, OSError):
+                    pass  # Value/IOError indicates negative sleep time in Py 3.5/2.7, respectively
+                allowance = rate_limit[0]
```
opaduchak, line 43: Question:
Do we need to ratelimit token refresh because don't want to overwhelm providers?

If so, maybe we can paralellize providers (refresh box, google-drive and mendeley at the same time, but do rate-limiting within each provider)

opaduchak (approved):

brianjgeiger (approved):

brianjgeiger (approved):

brianjgeiger (approved):

brianjgeiger (approved):

mfraezz (approved):
Will require accompanying helm chart changes from my end, but with a slight schedule adjustment, LGTM

PR #223, opened 2025-02-04
Only refresh token if the token have expired when returning wb creds

adlius:
We suspect that constant calls to external services to refresh tokens is behind the reason why performance is slow on staging and test. So we should only refresh tokens when we need to.


opaduchak (approved):
LGTM

PR #227, opened 2025-02-08
[ENG-7755] Enable session sharing between GV and OSF

adlius:


addon_service/authentication.py
```diff
@@ -12,7 +12,8 @@ def authenticate(self, request: DrfRequest):
         _user_uri = osf.get_osf_user_uri(request)
         if _user_uri:
             UserReference.objects.get_or_create(user_uri=_user_uri)
-            request.session["user_reference_uri"] = _user_uri
+            if not request.session.get("user_reference_uri"):
+                request.session["user_reference_uri"] = _user_uri
```
brianjgeiger, line 16: Do you need to get_or_create the UserReference if it's already on the session? Might be able to save a db call here if not.

brianjgeiger (approved):
One potential optimization, but overall looks good to me.

addon_service/authentication.py
brianjgeiger: Probably you still do need to do that, as I think about it more.

adlius: I tried to run the scenario in my mind, I think we do. Here a potential scenario immediately after deployment to prod:
- A logged out user logs into OSF, `user_reference_uri` is set on the shared session
- User (now authenticated) then makes a call to GV, at this point a `UserReference` instance is not yet created for this user on the GV side

opaduchak (approved):

app/settings.py
```diff
@@ -66,7 +66,24 @@
 OSF_BASE_URL = env.OSF_BASE_URL.rstrip("/")
 OSF_API_BASE_URL = env.OSF_API_BASE_URL.rstrip("/")
 ALLOWED_RESOURCE_URI_PREFIXES = {OSF_BASE_URL}
-SESSION_COOKIE_DOMAIN = env.SESSION_COOKIE_DOMAIN
+# SESSION_COOKIE_DOMAIN = env.SESSION_COOKIE_DOMAIN
```
cslzchen, line 69: I couldn't find `SESSION_COOKIE_DOMAIN` in the code base, so maybe we can simply remove it?

app/settings.py
```diff
@@ -66,7 +66,24 @@
...
+SESSION_ENGINE = "django.contrib.sessions.backends.cache"
+SESSION_COOKIE_NAME = env.OSF_AUTH_COOKIE_NAME
+OSF_AUTH_COOKIE_SECRET = env.OSF_AUTH_COOKIE_SECRET
+REDIS_HOST = env.REDIS_HOST
+
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
```
cslzchen, line 77: So, what's the difference between `django_redis.cache.RedisCache` (which OSF uses) and `django.core.cache.backends.redis.RedisCache` and why we choose the latter for GV?

app/settings.py
```diff
@@ -66,7 +66,24 @@
...
+        "LOCATION": REDIS_HOST,
+        # 'OPTIONS': {
+        #     'CLIENT_CLASS': 'django_redis.client.DefaultClient',
+        #     'CONNECTION_POOL_KWARGS': {
+        #         'max_connections': 100,
+        #     },
+        # },
```
cslzchen, line 84: Did we comment this out because this configuration doesn't work with `django.core.cache.backends.redis.RedisCache`? If so, we should just remove the code.

addon_service/common/osf.py
```diff
@@ -58,12 +58,7 @@ async def get_osf_user_uri(request: django_http.HttpRequest) -> str | None:
     _auth_headers = _get_osf_auth_headers(request)
     if not _auth_headers:
         return None
-    _client = await get_singleton_client_session()
-    async with _client.get(_osfapi_me_url(), headers=_auth_headers) as _response:
-        if HTTPStatus(_response.status).is_client_error:
-            return None
-        _response_content = await _response.json()
-        return _iri_from_osfapi_resource(_response_content["data"])
+    return request.session["user_reference_uri"]
```
cslzchen, line 61: I see. [Here](https://github.com/CenterForOpenScience/osf.io/pull/10965/files#diff-eb9961a790efef49b6d3b1039635e19bd3e21d0664c88df3318ffde221a49afeR38) is the OSF piece that support your change:

> 'user_reference_uri': user.get_semantic_iri(),

One question, should we replace this piece as well?

```python
_auth_headers = _get_osf_auth_headers(request)
if not _auth_headers:
    return None
````

e.g. just check if `request.session` has a valid `user_reference_uri`?

Finally, looking at `_get_osf_auth_headers()`, does this session stuff break requests using personal access token?

cslzchen (commented):
Looks good overall 🎆  with a few questions❓

app/settings.py
adlius: `django.core.cache.backends.redis.RedisCache` is the native Django `redis` adapter while `django_redis.cache.RedisCache` is an addon called `django_redis`. I believe when we were refactoring sessions for OSF we use the `django_redis` addon because we thought it'd gives us some features that improves performance. I chose to go with the native one as we don't really need `djang_redis`.

cslzchen (approved):
Changes looks good! While I am doing the integration testing, can you bring the PR up-to-date and also take a look at the failing tests? Thanks!

cslzchen (approved):
🌟

PR #229, opened 2025-02-10
fixed shared session issues and hmac auth

opaduchak:


addon_service/common/osf.py
```diff
@@ -48,17 +48,31 @@ def for_capabilities(capabilities: AddonCapabilities) -> "OSFPermission":
...
-    return request.session["user_reference_uri"]
+    # we must handle the case when old, already created sessions must somehow set user_reference_uri
+    # TODO: delete after all existing osf sessions have been revoked/expired
+    _client = await get_singleton_client_session()
+    async with _client.get(_osfapi_me_url(), headers=_auth_headers) as _response:
+        if HTTPStatus(_response.status).is_client_error:
+            return None
+        _response_content = await _response.json()
+        uri = _iri_from_osfapi_resource(_response_content["data"])
+        request.session["user_reference_uri"] = uri
```
adlius, line 75: I thought about this case and was actually thinking of asking Matt to log out all users upon release. But I suppose we can add this bit in just in case.

PR #237, opened 2025-02-14
[ENG-7171] Enable HMAC query of configured-storage-addons endpoint using pk in `<guid>:<addon_name>` format

adlius:


addon_service/configured_addon/storage/views.py
```diff
@@ -21,23 +23,35 @@ class ConfiguredStorageAddonViewSet(ConfiguredAddonViewSet):
     )
     serializer_class = ConfiguredStorageAddonSerializer
 
+    def get_object(self):
+        pk = self.kwargs["pk"]
+        if ":" in pk:
+            # Make sure only a valid HMAC request can query this endpoint using
+            # pk in "<guid>:<addon_name>" format
+            IsValidHMACSignedRequest().has_permission(self.request, self)
```
opaduchak, line 31: Are there any security concerns behind this?
Because I can imagine frontend also taking benefit from this lookup in the future

opaduchak (approved):
Looks great!

addon_service/configured_addon/storage/views.py
adlius: Not a security concern but I believe this is against the design philosophy of GV. I believe we choose to use opaque ids for configured addons, instead of using guids from OSF, is because in this way we make no assumptions about the relationship between OSF entities and GV addons. But if we rely on querying configured addons in `<guid>:<addon_name>` format, the assumption is that there is *one* configured addon of the same external service per OSF node. I wanna make sure we don't rely on this pattern cuz I don't think this is going to be true moving forwards.

PR #239, opened 2025-02-21
[ENG-7290] added management comand for easy local provider setup

opaduchak:
Motivation: Provider setup may be tedious process, so to mitigate this, management command has been added, which fills all existing providers with exemption for secrets to have an ability to roll development environment quickly.
Currently it supports only local env, if needed it may be expanded to prod and staging envs


sh-andriy (approved):
LGTM!

PR #245, opened 2025-02-28
[ENG-7319] | Added `try except` to print out full error message if the issue persists

sh-andriy:
**`[ENG-7319]`** | Added `try except` to print out full error message if the issue persists

[ENG-7319]: https://openscience.atlassian.net/browse/ENG-7319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/addon_imps/storage/dropbox.py
+++ b/addon_imps/storage/dropbox.py
@@ -37,6 +37,12 @@ class DropboxStorageImp(storage.StorageAddonHttpRequestorImp):
                 "path": self.config.connected_root_id,
             },
         ) as _response:
+            if _response.http_status != 200:
+                error_content = await _response.text_content()
+                raise ValueError(
+                    f"Request failed with status {_response.http_status}: {error_content}"
+                )
+
             content = await _response.json_content()
 
             if "error" in content:
--- a/addon_imps/storage/dropbox.py
+++ b/addon_imps/storage/dropbox.py
@@ -38,9 +38,14 @@ class DropboxStorageImp(storage.StorageAddonHttpRequestorImp):
             },
         ) as _response:
             content = await _response.json_content()
-            return {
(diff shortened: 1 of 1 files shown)


addon_imps/storage/dropbox.py
```diff
@@ -38,9 +38,14 @@ async def build_wb_config(self) -> dict:
             },
         ) as _response:
             content = await _response.json_content()
-            return {
-                "folder": content["path_display"],
-            }
+
+            if "error" in content:
+                raise ValueError(f"Dropbox API error: {content}")
```
opaduchak, line 43: Nit: maybe also check by status code being not 200?

addon_imps/storage/dropbox.py
```diff
@@ -38,9 +38,14 @@ async def build_wb_config(self) -> dict:
             },
         ) as _response:
             content = await _response.json_content()
-            return {
-                "folder": content["path_display"],
-            }
+
+            if "error" in content:
```
adlius, line 42: The `error` key would appear in `content` only if there is an error, correct?

adlius (commented):
A question. Otherwise looks good.

addon_imps/storage/dropbox.py
sh-andriy: Yes, the `error` key in content would only appear if there is an `error` in the response from **`Dropbox`**.

addon_imps/storage/dropbox.py
sh-andriy: Sure, I think it would be a good improvement

opaduchak (approved):

PR #247, opened 2025-03-03
avoid N+1 request when display projects' guid related to each AuthorizedStorageAccount

ihorsokhanexoft:
As we want to display projects' guid for each authorized addon account, it'll create N+1 issue, thus using configured_storage_addons as ResourceRelatedField would be inefficient plus creates other issues whose solution wouldn't follow JSONAPI approach

### DIFF
--- a/addon_service/authorized_account/storage/serializers.py
+++ b/addon_service/authorized_account/storage/serializers.py
@@ -57,6 +57,16 @@ class AuthorizedStorageAccountSerializer(AuthorizedAccountSerializer):
         "authorized_operations": "addon_service.serializers.AddonOperationSerializer",
     }
 
+    configured_storage_addons_uris = serializers.SerializerMethodField()
+
+    def get_configured_storage_addons_uris(self, obj):
+        return obj.configured_storage_addons.select_related(
+            'authorized_resource'
+        ).values_list(
+            'authorized_resource__resource_uri',
+            flat=True
+        )
+
     class Meta:
         model = AuthorizedStorageAccount
         fields = [
@@ -75,4 +85,5 @@ class AuthorizedStorageAccountSerializer(AuthorizedAccountSerializer):
             "external_storage_service",
             "initiate_oauth",
             "credentials_available",
+            "configured_storage_addons_uris",
         ]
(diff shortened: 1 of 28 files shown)


addon_service/authorized_account/storage/serializers.py
```diff
@@ -57,6 +57,16 @@ class AuthorizedStorageAccountSerializer(AuthorizedAccountSerializer):
...
 
+    configured_storage_addons_uris = serializers.SerializerMethodField()
+
+    def get_configured_storage_addons_uris(self, obj):
+        return obj.configured_storage_addons.select_related(
+            'authorized_resource'
+        ).values_list(
+            'authorized_resource__resource_uri',
+            flat=True
+        )
```
opaduchak, line 68: this should be defined also for `AuthorizedCitationAccount` and `AuthorizedComputingAccount`

opaduchak (changes requested):
Looks great, please also add this attribute to citation and compute accounts

ihorsokhanexoft:
@opaduchak I noticed that `configured_storage_addons` doesn't use .active() method. Is it correct for other addons account?

opaduchak:
> @opaduchak I noticed that `configured_storage_addons` doesn't use .active() method. Is it correct for other addons accounts?

I believe so, because we are already fetching active accounts

ihorsokhanexoft:
> > @opaduchak I noticed that `configured_storage_addons` doesn't use .active() method. Is it correct for other addons accounts?
> 
> I believe so, because we are already fetching active accounts

So, no need to use .active(), right?

opaduchak (approved):
Looks great!
I have one small sugestion, please do it if you have the capacity

addon_service/user_reference/views.py
```diff
@@ -11,5 +11,6 @@ class UserReferenceViewSet(RestrictedReadOnlyViewSet):
     permission_classes = [
         SessionUserIsOwner,
     ]
+    allowed_query_params = ['uris']
```
opaduchak, line 14: may you also do this for `ViewOnlyFilter` as well, please?

opaduchak (approved):

opaduchak (approved):
LGTM!

PR #248, opened 2025-03-04
[ENG-7345] added management command to merge authorized account credentials

opaduchak:
**TL;DR** `credentials_format` will be stored in `ExternalCredentials` to make this fix work

There are some controversial decisions to make this fix work:
We need `AuthorizedAccount`->`ExternalCredentials` to be Many2One relationship instead of One2One. This change has caused some side effects:

Backpopulated field `authorized_account` on `ExternalCredentials` stopped working (it would return empty list instead of respective accounts) till both `ExternalCredentials` `AuthorizedAccount` are saved, which causes issues when creating new `AuthorizedAccount` because this field is needed to perform `full_clean`, which is performed before any `save`.

Based on that, I've decided to migrate `credential_format` to `ExternalCredentials` which will reduce query count and will fix this scenario.


adlius:
@opaduchak Looks like there is not a convenient way for me to reopen

addon_service/management/commands/merge_migrated_accounts.py
```diff
@@ -0,0 +1,81 @@
...
+    "mendeley": "citations",
+}
+
+
+def get_class(integration_type):
+    if integration_type == "storage":
+        return AuthorizedStorageAccount
+    elif integration_type == "citations":
+        return AuthorizedCitationAccount
+    elif integration_type == "computing":
```
adlius, line 38: Probably don't need to worry about computing so far as we have not migrated any Boa accounts

adlius (approved):
One comment. But otherwise looks good.

addon_service/management/commands/merge_migrated_accounts.py
```diff
@@ -0,0 +1,81 @@
...
+                )
+                .select_related("oauth2_token_metadata", "_credentials")
+                .order_by(
+                    F("oauth2_token_metadata__date_last_refreshed").desc(
+                        nulls_last=True
+                    )
+                )
+            )
+            # take the first account here because it has the most recently refreshed credentials,
+            # and update all credentials/token metadata with first_account's ones
```
mfraezz, line 74: Not strictly required, but is there any simple way to confirm that a given set of credentials is valid? i.e. some non-destructive request to issue against the service to confirm it returns a `20X` rather than a `40X`

addon_service/management/commands/migrate_authorized_account.py
```diff
@@ -233,55 +233,59 @@ def migrate_for_user(
...
+        account_map = {}
+        for osf_account in users_external_accounts:
+            try:
+                credentials = self.get_credentials(external_service, osf_account)
+            except CredentialException as e:
+                logger.error(
+                    f"Skipping account migration with {osf_account.pk=} for service {external_service.display_name} "
+                    f"due to credentials parse error {e=}"
+                )
+                return
```
mfraezz, line 246: This should likely be `continue` now that it may need to check multiple `ExternalAccount`s.

addon_service/management/commands/migrate_authorized_account.py
```diff
@@ -233,55 +233,59 @@ def migrate_for_user(
...
+            account.save(full_clean=False)
+            account_map[osf_account.id] = account
 
         for node_settings in node_settings_set:
+            node_guid = get_node_guid(node_settings.owner_id)
             resource_reference = ResourceReference.objects.get_or_create(
-                resource_uri=f"{OSF_BASE}/{get_node_guid(node_settings.owner_id)}"
+                resource_uri=f"{OSF_BASE}/{node_guid}"
             )[0]
+            base_account = account_map.get(node_settings.external_account.id)
```
mfraezz, line 283: May need to `continue` here `if not base_account`, as a consequence of any `CredentialException` above

addon_service/management/commands/merge_migrated_accounts.py
```diff
@@ -0,0 +1,81 @@
...
+            token_metadata = accounts[0].oauth2_token_metadata
+            credentials = accounts[0]._credentials
+            accounts.update(
```
mfraezz, line 77: Would it be possible to log relevant info in here? -- Any account / credential identifiers which could help to understand what changes, and to potentially allow an individual change to be reverted if it breaks.

mfraezz (commented):
Generally looks good, with one question, a verbosity request, and a couple notes regarding edge cases.

addon_service/credentials/models.py
```diff
@@ -99,16 +109,20 @@ def authorized_accounts(self):
                 *filter(
```
mfraezz, line 109: Minor: docstring for this function could be updated, as it's not longer accurate

mfraezz:
Aside: I didn't find evidence of it in the code, but if we intentionally invalidate or delete `ExternalCredentials` when, for example, if a user deauthorizes an account, we'll additionally need to ensure those creds are not shared with any other `AuthorizedAccount` first.

addon_service/management/commands/merge_migrated_accounts.py
opaduchak: We may try to construct something in place, but this won't be simple

opaduchak:
> Aside: I didn't find evidence of it in the code, but if we intentionally invalidate or delete `ExternalCredentials` when, for example, if a user deauthorizes an account, we'll additionally need to ensure those creds are not shared with any other `AuthorizedAccount` first.

We don't work with `ExternalCredentials` directly, all of this is handled by `AuthorizedAccount`, consequently `ExternalCredentials` may be deleted only by cascade, so this edgecase should not happen except some manual intervensions

addon_service/management/commands/merge_migrated_accounts.py
mfraezz: This heuristic is probably fine, then.

addon_service/migrations/0013_externalcredentials_int_credentials_format_and_more.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+import addon_service.common.validators
+from addon_service.credentials.models import ExternalCredentials
+
+
+def migrate_credential_format(*args, **kwargs):
+    skipped = 0
+    processed = 0
+    for credential in ExternalCredentials.objects.iterator():
```
mfraezz, line 15: Is it possible to optimize this, e.g. with `.filter(<...>).update(<...>)`?

Not done benchmarking it locally yet, but a rough estimate for this runtime is a little over an hour for prod.

addon_service/management/commands/merge_migrated_accounts.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+            # take the first account here because it has the most recently refreshed credentials,
+            # and update all credentials/token metadata with first_account's ones
+            account = accounts[0]
+            token_metadata = accounts.oauth2_token_metadata
+            credentials = accounts._credentials
+            logger.info(
+                f"Merging accounts that point to {account.external_service.external_service_name} with id {account.id}"
+            )
+            accounts.update(
```
mfraezz, line 83:
```suggestion
            live_account = accounts[0]
            token_metadata = live_account.oauth2_token_metadata
            credentials = live_account._credentials
            logger.info(
                f"Merging shared accounts for {account.external_service.external_service_name}:{account.provider_id} with id {live_account.id}"
            )
            logger.info(f"Affected accounts: {accounts.values_list('id', flat=True)}")
            accounts.update(
```

addon_service/management/commands/merge_migrated_accounts.py
```diff
@@ -0,0 +1,87 @@
...
+                )
+            )
+            # take the first account here because it has the most recently refreshed credentials,
+            # and update all credentials/token metadata with first_account's ones
+            account = accounts[0]
+            token_metadata = account.oauth2_token_metadata
+            credentials = account._credentials
+            logger.info(
+                f"Merging accounts that point to {account.external_service.external_service_name} with id {account.id}"
+            )
```
mfraezz, line 82: Need to ensure non-zero `accounts.count()` to avoid a `KeyError` (`>1` to avoid a no-op), and `info` logs get hidden by default `log_level`. Added a few extra logs to specify what changes are being made.

```suggestion
            if accounts.count() < 2:
                logger.warn(f"Found {accounts.count()} accounts for {account.provider} {account.provider_id}, skipping")
                continue
            # take the first account here because it has the most recently refreshed credentials,
            # and update all credentials/token metadata with first_account's ones
            account = accounts[0]
            token_metadata = account.oauth2_token_metadata
            credentials = account._credentials
            logger.warn(
                f"Merging accounts that point to {account.external_service.external_service_name} with id {account.id}"
            )
            for account_to_update in accounts:
                last_refresh = getattr(account_to_update.oauth2_token_metadata, 'date_last_refreshed', 'NO_TOKEN_META')
                if account_to_update.id == account.id:
                    logger.warn(f'Using credentials from {account_to_update.id} refreshed on {last_refresh}')
                else:
                    logger.warn(f'Updating {account_to_update.id} refreshed on {last_refresh}')
```

mfraezz (approved):

PR #249, opened 2025-03-06
[ENG-7344] Fix migration for users with multiple ExternalAccount to the same provider

adlius:
This is to fix the previous migration mistake where we only used the first ExternalAccount of the same provider for the OSFUser and apply it to all ConfiguredAddons (the equivalent of NodeSettings on OSF). The fix is to first find instances where a user have multiple ExternalAccount for the same provider. For each user-provider pair, we find all the NodeSettings and find their respective migrated ConfiguredAddons in GV and check to see if the ConfiguredAddon's external_account_id matches the NodeSettings' provider_id. If they match, that means they are migrated correctly and we do nothing. If not, then we fix them.

### DIFF
--- a/addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
+++ b/addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
@@ -1,0 +1,271 @@
+import logging
+
+from django.contrib.contenttypes.models import ContentType
+from django.core.cache import cache
+from django.core.management import BaseCommand
+from django.db import transaction
+from django.db.models import Count
+
+from addon_service.authorized_account.citation.models import AuthorizedCitationAccount
+from addon_service.authorized_account.storage.models import AuthorizedStorageAccount
+from addon_service.common.credentials_formats import CredentialsFormats
+from addon_service.configured_addon.citation.models import ConfiguredCitationAddon
+from addon_service.configured_addon.storage.models import ConfiguredStorageAddon
+from addon_service.external_service.models import ExternalService
+from addon_service.osf_models.models import (
+    BitbucketUserSettings,
+    BoxUserSettings,
+    DataverseUserSettings,
+    DropboxUserSettings,
+    ExternalAccount,
+    FigshareUserSettings,
+    GithubUserSettings,
+    GitlabUserSettings,
+    GoogleDriveUserSettings,
+    Guid,
+    MendeleyUserSettings,
+    OneDriveUserSettings,
+    OsfUser,
+    OwnCloudUserSettings,
+    S3UserSettings,
+    ZoteroUserSettings,
+)
+from addon_service.resource_reference.models import ResourceReference
+from addon_service.user_reference.models import UserReference
+from addon_toolkit import AddonCapabilities
+from addon_toolkit.credentials import (
+    AccessKeySecretKeyCredentials,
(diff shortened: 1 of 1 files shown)


opaduchak (approved):

addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
```diff
@@ -0,0 +1,271 @@
...
+
+    external_service = get_external_service(provider)
+    account_owner = get_user_reference(user_guid)
+
+    authorized_account = None
+    if not AuthorizedAccount.objects.filter(
+        external_account_id=external_account.id,
+        external_service=external_service,
+        account_owner=account_owner,
+    ).exists():
```
mfraezz, line 203: Could this run into the same issue where multiple users have access to the same `ExternalAccount`? 

For example, given Users A and B, and ExternalAccounts X, Y, Z: Assuming A had X, Y connected, and B had Y, Z (in that order), the initial migration would've migrated both accounts X and Y to users A and B, respectively. However, I think this command would migrate Y again with old credentials for User A, which may have been updated more recently in GV's DB based on User B's behavior.

If I'm right about that, may need to remove `account_owner` from the initial filter, and then handle that possible case as well.

addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
```diff
@@ -0,0 +1,271 @@
...
+
+def get_external_service(provider):
+    return ExternalService.objects.get(wb_key=provider)
+
+
+def get_user_reference(user_guid):
+    return UserReference.objects.get_or_create(user_uri=f"{OSF_BASE}/{user_guid}")
+
+
+def get_or_create_authorized_account(external_account, provider, user_guid):
```
mfraezz, line 189: Could you please add some logging in this function to say what actions it's performing, e.g. `Creating AuthorizedAccount on {provider} for {user}`, `Found AuthorizedAccount {id} for {user}`, `Found shared credentials on {provider} for users {users_0...n}`, etc?

addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
```diff
@@ -0,0 +1,271 @@
...
+        # check to see if the migrated configured_addon instance have the authorized account with the same external_account_id
+        if not configured_addon_has_correct_base_account(
+            external_account, node_guid, provider
+        ):
+            # if not, then we fix it
+            configured_addon = get_configured_addon(node_guid, provider)
+            authorized_account = get_or_create_authorized_account(
+                external_account, provider, user_guid
+            )
+            configured_addon.base_account = authorized_account
```
mfraezz, line 254: Could you please add some logging in here as well, to track what's being wired up?

mfraezz (commented):
Generally looks good. One potential edge case, a couple requests for verbosity.

addon_service/management/commands/fix_migrating_users_with_mutiple_external_accounts.py
adlius: @mfraezz I think this edge case would be handled by #248 . The idea is to run this fix first and then the fix in #248 which is implemented to fix specifically the issue that you mentioned. Do you think it would be sufficient?

PR #255, opened 2025-03-28
[ENG-7568][ENG-7569][ENG-7570][ENG-7571] Added external_service, configured_addon, authorized_account

opaduchak:


addon_service/external_service/link/models.py
```diff
@@ -0,0 +1,65 @@
...
+    ADD_UPDATE_FILES_PARTIAL = auto()
+    DELETE_FILES = auto()
+    DELETE_FILES_PARTIAL = auto()
+    FORKING = auto()
+    LOGS = auto()
+    PERMISSIONS = auto()
+    REGISTERING = auto()
+    FILE_VERSIONS = auto()
+    COPY_INTO = auto()
+    DOWNLOAD_AS_ZIP = auto()
```
opaduchak, line 27: This is a placeholder till we implement true ResourceType

addon_toolkit/interfaces/link.py
```diff
@@ -0,0 +1,111 @@
...
+    "ItemType",
+    "ItemSampleResult",
+    "LinkAddonInterface",
+    "LinkAddonImp",
+)
+
+
+class ItemType(enum.StrEnum):
+    FILE = enum.auto()
+    FOLDER = enum.auto()
```
opaduchak, line 29: I guess this will not be replaced by resource type, because we will still need to traverse the tree

addon_service/admin/__init__.py
```diff
@@ -22,7 +23,7 @@ class ExternalStorageServiceAdmin(GravyvaletModelAdmin):
     )
     raw_id_fields = ("oauth2_client_config", "oauth1_client_config")
     enum_choice_fields = {
-        "int_addon_imp": known_imps.AddonImpNumbers,
+        "int_addon_imp": known_imps.StorageAddonImpNumbers,
```
opaduchak, line 26: this should make admin setup easier

addon_service/admin/_base.py
```diff
@@ -56,10 +56,7 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
                 label=db_field.verbose_name,
                 choices=[
                     (None, ""),
-                    *(
-                        (_member.value, _member.name)
-                        for _member in _enum.__members__.values()
-                    ),
+                    *((_member.value, _member.name) for _member in _enum),
```
opaduchak, line 59: this is to accomodate easier admin setup, moreover it is simpler and doesn't break anything

addon_service/configured_addon/link/models.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+
+from addon_service.configured_addon.models import ConfiguredAddon
+
+
+class ConfiguredLinkAddon(ConfiguredAddon):
+
+    target_uri = models.URLField()
+    target_id = models.CharField()
+    int_resource_type = models.IntegerField()
```
opaduchak, line 10: AFAIK we don't create configured addon without these attributes. Maybe we don't need to allow them blank?

addon_service/addon_operation_invocation/views.py
```diff
@@ -78,6 +84,10 @@ def retrieve_related(self, request, *args, **kwargs):
                 serializer = ConfiguredComputingAddonSerializer(
                     instance, context={"request": request}
                 )
+            elif hasattr(instance, "configuredcomputingaddon"):
```
adlius, line 87:
```suggestion
            elif hasattr(instance, "configuredlinkaddon"):
```
Typo?

addon_service/admin/__init__.py
```diff
@@ -50,6 +51,25 @@ class ExternalCitationServiceAdmin(GravyvaletModelAdmin):
...
+class ExternalLinkServiceAdmin(GravyvaletModelAdmin):
+    list_display = ("display_name", "created", "modified")
+    readonly_fields = (
+        "id",
+        "created",
+        "modified",
+    )
+    raw_id_fields = ("oauth2_client_config", "oauth1_client_config")
+    enum_choice_fields = {
+        "int_addon_imp": known_imps.AddonImpNumbers,
```
adlius, line 64: Should there be a `known_imps.LinkAddonImpNumbers` and make use of it here?

addon_service/authorized_account/link/serializers.py
```diff
@@ -0,0 +1,86 @@
...
+)
+from addon_service.models import (
+    AuthorizedStorageAccount,
+    ConfiguredLinkAddon,
+    ExternalLinkService,
+    UserReference,
+)
+
+
+RESOURCE_TYPE = get_resource_type_from_model(AuthorizedStorageAccount)
```
adlius, line 23:
```suggestion
RESOURCE_TYPE = get_resource_type_from_model(AuthorizedLinkAccount)
```
Typo?

addon_service/common/known_imps.py
```diff
@@ -114,3 +114,22 @@ class AddonImpNumbers(enum.Enum):
...
+        AddonImpNumbers.ONEDRIVE,
+        AddonImpNumbers.OWNCLOUD,
+        AddonImpNumbers.DATAVERSE,
+        AddonImpNumbers.GITLAB,
+        AddonImpNumbers.BITBUCKET,
+        AddonImpNumbers.GITHUB,
+    }
+)
+CitationAddonImpNumbers = frozenset({AddonImpNumbers.ZOTERO, AddonImpNumbers.MENDELEY})
+ComputingAddonImpNumbers = frozenset({AddonImpNumbers.BOA})
```
adlius, line 135: Maybe we should add a `LinkAddonImpNumbers` here?

addon_service/configured_addon/link/models.py
```diff
@@ -0,0 +1,33 @@
...
+        if len(resource_type) != 1:
+            raise ValidationError("One may select only one resource type")
+    except ValueError:
+        raise ValidationError("Invalid resource type")
+
+
+class ConfiguredLinkAddon(ConfiguredAddon):
+
+    target_uri = models.URLField()
+    target_id = models.CharField()
```
adlius, line 20: Do we need both `target_uri` and `target_id`? Do we plan to serialize `target_uri` directly as the link of the `ConfiguredLinkAddon`? I thought we talked about having a getter that'd get the link based on the `target_id` and we would serialize that getter.

addon_service/configured_addon/link/serializers.py
```diff
@@ -0,0 +1,82 @@
...
+from addon_toolkit.interfaces.link import SupportedResourceTypes
+
+
+RESOURCE_TYPE = get_resource_type_from_model(ConfiguredLinkAddon)
+
+
+class ConfiguredLinkAddonSerializer(ConfiguredAddonSerializer):
+    """api serializer for the `ConfiguredLinkAddon` model"""
+
+    target_uri = URLField()
```
adlius, line 27: Again, following the question above, if `target_uri` is the actual link of the `ConfiguredLinkAddon`, are we asking the FE to set this `target_uri`? I thought the idea is to have the BE serialize the `target_uri` based on the `target_id` that was set by the FE.

adlius (changes requested):
Some questions.

addon_service/admin/__init__.py
opaduchak: Yeah, but don't have them yet

addon_imps/link/dataverse.py
```diff
@@ -0,0 +1,195 @@
...
+class DataverseLinkImp(LinkAddonHttpRequestorImp):
+    """storage on dataverse
+
+    see https://guides.dataverse.org/en/latest/api/native-api.html
+    """
+
+    async def build_url_for_id(self, item_id: str) -> str:
+        dataset_id = DATASET_REGEX.match(item_id)["id"]
+        dataset = await self._fetch_dataset(dataset_id)
+        return dataset.item_id
```
opaduchak, line 31: this is a dummy implementation for now, but it seems that we will still need config for link addons after all

addon_service/common/known_imps.py
opaduchak: Fixed, also I've made them autogenerated

addon_service/configured_addon/link/models.py
opaduchak: FIxed this to be generated from addon.
I've added this as a field because I thought that this would be easier to work with before we implement link addons

addon_service/configured_addon/link/serializers.py
opaduchak: This will remain but now it will serialize link from addon, also, I've made this read-only field

PR #261, opened 2025-04-09
[ENG-7757][ENG-7310] 3.2.3 GV: Enable GV->OSF messaging - (GV part)

opaduchak:


addon_service/configured_addon/models.py
```diff
@@ -88,6 +89,20 @@ def connected_operations(self) -> list[AddonOperationModel]:
             )
         ]
 
+    def save(self, *args, full_clean=True, **kwargs):
+        id_ = self.pk
+        super().save(*args, full_clean=full_clean, **kwargs)
+        if not id_:  # If instance is created, not updated
+            app.send_task(
```
cslzchen, line 96: So correct me if I am wrong, is this `.send_task()` the one that forces us to configure `rabbitmq` when running tests? Normally, I don't think we need `rabbitmq` if just testing tasks.

cslzchen (approved):
Looks good with a minor non-blocking question 🎆

cslzchen (approved):
🎆

PR #262, opened 2025-04-11
[ENG-7777] [ENG-7778] 2.2 GV: Add a new LINK addon implementation: Dataverse

opaduchak:
## Ticket

* https://openscience.atlassian.net/browse/ENG-7777
* https://openscience.atlassian.net/browse/ENG-7778


addon_toolkit/interfaces/link.py
```diff
@@ -71,6 +71,12 @@ class ItemResult:
     item_type: ItemType
     resource_type: SupportedResourceTypes | None = None
     item_link: str | None = None
+    doi: str | None = None
+
+
+@dataclasses.dataclass(frozen=True)
+class LinkConfig:
+    external_api_url: str
```
opaduchak, line 79: It seems that we need config after all, because we need to create verified link with `api_base_url`

addon_service/authorized_account/link/models.py
```diff
@@ -22,3 +23,9 @@ def configured_link_addons(self):
         return ConfiguredLinkAddon.objects.filter(base_account=self).select_related(
             "authorized_resource"
         )
+
+    @property
+    def config(self) -> LinkConfig:
+        return LinkConfig(
+            external_api_url=self.api_base_url,
+        )
```
cslzchen, line 31: Need to update this if my other comments make sense.

cslzchen (changes requested):
Looks great 🌟 with one question on API vs Web URLs.

cslzchen (approved):
Looks good and I will do a quick local integration test before merging it.

addon_imps/link/dataverse.py
```diff
@@ -52,17 +64,11 @@ async def list_root_items(self, page_cursor: str = "") -> ItemSampleResult:
...
-                        "Draft",
-                        "Deaccessioned",
-                        "In+Review",
-                    ]
-                ],
+                (
+                    "dvobject_types",
+                    "Dataverse",
+                ),  # only published dataverses may contain published datasets
+                ("published_states", "Published"),
```
adlius, line 71: So we only allow users to link to objects that is `Published` on dataverse? Is that a requirement?

adlius (approved):
Looks good. One minor question.

addon_imps/link/dataverse.py
opaduchak: Yes,as I understand this is a requirement, because otherwise link will be private

cslzchen: Yes, as Oleh has mentioned, the link needs to be a public link with public resources.

PR #272, opened 2025-04-28
[ENG-7755] Bug-fix: Final fix for PAT

opaduchak:
Final fix for PAT:
- Found all instances of accessing user_reference_uri from session, replaced them with helper function call
- extracted getting user uri into helper function


cslzchen (approved):
⭐

PR #276, opened 2025-04-30
[Reference PR] VRL - GV Part

cslzchen:


app/env.py
```diff
@@ -46,6 +47,7 @@
 OSF_BASE_URL = os.environ.get("OSF_BASE_URL", "https://osf.example")
 OSF_API_BASE_URL = os.environ.get("OSF_API_BASE_URL", "https://api.osf.example")
 OSF_AUTH_COOKIE_NAME = os.environ.get("OSF_AUTH_COOKIE_NAME", "osf_staging")
+OSF_AUTH_COOKIE_SECRET = os.environ.get("OSF_AUTH_COOKIE_SECRET", "CHANGEME")
```
cslzchen, line 50: GV Deployment Note:

* `OSF_AUTH_COOKIE_NAME` should be the same as what OSF uses in `COOKIE_NAME`
* `OSF_AUTH_COOKIE_SECRET` should be the same as what OSF uses in `SECRET_KEY`

app/settings.py
```diff
@@ -62,11 +62,21 @@
...
-USER_REFERENCE_COOKIE = env.OSF_AUTH_COOKIE_NAME
+OSF_AUTH_COOKIE_NAME = env.OSF_AUTH_COOKIE_NAME
 OSF_BASE_URL = env.OSF_BASE_URL.rstrip("/")
 OSF_API_BASE_URL = env.OSF_API_BASE_URL.rstrip("/")
 ALLOWED_RESOURCE_URI_PREFIXES = {OSF_BASE_URL}
-SESSION_COOKIE_DOMAIN = env.SESSION_COOKIE_DOMAIN
+SESSION_ENGINE = "django.contrib.sessions.backends.cache"
+SESSION_COOKIE_NAME = env.OSF_AUTH_COOKIE_NAME
+OSF_AUTH_COOKIE_SECRET = env.OSF_AUTH_COOKIE_SECRET
+REDIS_HOST = env.REDIS_HOST
```
cslzchen, line 72: GV Deployment Note: should be good after we set the `env` variables

PR #278, opened 2025-05-05
[ENG-7932] GV: Ease restrictions for `resource_type` and `target_id`

cslzchen:
## Purpose

Ease restrictions for `resource_type` and `target_id` for `ConfiguredLinkAddon`

## Side Effect

Ideally, we don't need to ease the restriction. However, FE flow has been wired in a way that a temporary `ConfiguredLinkAddon` instance must be saved in before we can continue to update `resource_type` and `target_id`. Re-wiring the FE workflow is out of the scope of this project so we had to ease the restrictions.

## Ticket

https://openscience.atlassian.net/browse/ENG-7932


addon_service/configured_addon/link/models.py
```diff
@@ -11,6 +11,8 @@
 
 
 def is_supported_resource_type(resource_type: int):
+    if not resource_type:
+        pass
```
adlius, line 15: Do you mean `return` ?

addon_service/configured_addon/link/models.py
```diff
@@ -20,34 +22,43 @@ def is_supported_resource_type(resource_type: int):
 
 
 class ConfiguredLinkAddon(ConfiguredAddon):
-    target_id = models.CharField()
-    int_resource_type = models.BigIntegerField(validators=[is_supported_resource_type])
+    target_id = models.CharField(
+        default=None, null=True
```
adlius, line 26: How about `null=True, blank=True`?

adlius (commented):
Some questions.

addon_service/configured_addon/link/models.py
cslzchen: Good eyes, yes

addon_service/configured_addon/link/models.py
cslzchen: Works for me

cslzchen: Update: I will change all to `blank=True` so we can do loose `Fasly` check instead of strict `None` Check

addon_service/configured_addon/link/models.py
```diff
@@ -20,34 +22,40 @@ def is_supported_resource_type(resource_type: int):
...
+    target_id = models.CharField(default=None, null=True, blank=True)
+    int_resource_type = models.BigIntegerField(
+        default=None, null=True, blank=True, validators=[is_supported_resource_type]
+    )
 
     @property
-    def resource_type(self) -> SupportedResourceTypes:
-        return SupportedResourceTypes(self.int_resource_type)
+    def resource_type(self) -> SupportedResourceTypes | None:
+        if not self.int_resource_type:
```
adlius, line 32:
```suggestion
        if self.int_resource_type:
```
I think there is a typo here

addon_service/configured_addon/link/serializers.py
```diff
@@ -25,9 +25,11 @@
...
     """api serializer for the `ConfiguredLinkAddon` model"""
 
-    target_url = URLField(read_only=True)
-    target_id = CharField()
-    resource_type = EnumNameChoiceField(enum_cls=SupportedResourceTypes)
+    target_url = URLField(allow_null=True, allow_blank=True, read_only=True)
+    target_id = CharField(allow_null=True, allow_blank=True)
+    resource_type = EnumNameChoiceField(
+        allow_null=True, allow_blank=True, enum_cls=SupportedResourceTypes
+    )
```
adlius, line 32:
```suggestion
    target_id = CharField(required=False)
    resource_type = EnumNameChoiceField(
        required=False,
        enum_cls=SupportedResourceTypes,
    )
```

adlius: I think adding `required=False` should do.

adlius (changes requested):
Some suggestions after checking it locally

addon_service/configured_addon/link/serializers.py
```diff
@@ -25,9 +25,12 @@
 class ConfiguredLinkAddonSerializer(ConfiguredAddonSerializer):
     """api serializer for the `ConfiguredLinkAddon` model"""
 
-    target_url = URLField(read_only=True)
-    target_id = CharField()
-    resource_type = EnumNameChoiceField(enum_cls=SupportedResourceTypes)
+    target_url = URLField(allow_null=True, allow_blank=True, read_only=True)
```
cslzchen, line 28: @adlius because `target_url` is read-only, do we still need `allow_null=True, allow_blank=True`? My understanding is that both options applies to writable fields only.

cslzchen: Ah, looked into it a little bit more, in class `Field` we have:

```
# If `required` is unset, then use `True` unless a default is provided.
if required is None:
    required = default is empty and not read_only

# Some combinations of keyword arguments do not make sense.
assert not (read_only and write_only), NOT_READ_ONLY_WRITE_ONLY
assert not (read_only and required), NOT_READ_ONLY_REQUIRED
assert not (required and default is not empty), NOT_REQUIRED_DEFAULT
assert not (read_only and self.__class__ == Field), USE_READONLYFIELD
```

adlius: DRF documents are somewhat vague about this. For example, it says [the following](https://www.django-rest-framework.org/api-guide/fields/#allow_null) for the `allow_null` arugment:

> Normally an error will be raised if None is passed to a serializer field. Set this keyword argument to True if None should be considered a valid value.

>Note that, without an explicit default, setting this argument to True will imply a default value of null for serialization output, but does not imply a default for input deserialization.

>Defaults to False

It reads to me that it applies to both serialization deserialization.

For `allow_blank` option of string fields, it says [the following](https://www.django-rest-framework.org/api-guide/fields/#string-fields):

> If set to True then the empty string should be considered a valid value. If set to False then the empty string is considered invalid and will raise a validation error. Defaults to False.

So it doesn't really say whether it's for serialization or deserialization or both. But if we follow the Django paradigm where `blank` determines whether a field can be blank on form submissions, then I'd wager that this only applies to deserialization. Interestingly, it also adds that

> The allow_null option is also available for string fields, although its usage is discouraged in favor of allow_blank. It is valid to set both allow_blank=True and allow_null=True, but doing so means that there will be two differing types of empty value permissible for string representations, which can lead to data inconsistencies and subtle application bugs.

So maybe we should skip both fields for this?

cslzchen: I think we can keep `allow_null=True` and `allow_blank=True` since they are checked in serializer getters.

cslzchen: Agreed and I have removed the fields.

adlius (approved):

adlius (approved):

PR #280, opened 2025-05-06
[ENG-7931] GV: Hybrid type missing in GV admin

opaduchak:
## Purpose

Fix the bug which causes `HYBRID` type to be missing from GV Admin.

## Ticket

https://openscience.atlassian.net/browse/ENG-7931

### DIFF
--- a/addon_service/admin/_base.py
+++ b/addon_service/admin/_base.py
@@ -56,7 +56,7 @@ class GravyvaletModelAdmin(admin.ModelAdmin):
                 label=db_field.verbose_name,
                 choices=[
                     (None, ""),
-                    *((_member.value, _member.name) for _member in _enum),
+                    *self.iter_enum_members(_enum),
                 ],
             )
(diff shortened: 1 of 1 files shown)


addon_service/admin/_base.py
```diff
@@ -65,11 +65,17 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
...
 
         return super().formfield_for_dbfield(db_field, request, **kwargs)
+
+    def _list_enum_members(self, _enum):
+        if hasattr(_enum, "__members__"):
+            iterator = _enum.__members__.values()
+        else:
+            iterator = _enum
+
+        return [(_member.value, _member.name) for _member in iterator]
```
cslzchen, line 81: Fix looks good but I am curious on what changes we had made that led to some `_enum` no longer has `__members__`.

cslzchen (approved):
⭐

PR #285, opened 2025-05-15
Reference Zenodo imp

opaduchak:
Reference implementation for Zenodo addon

### DIFF
--- a/addon_service/common/known_imps.py
+++ b/addon_service/common/known_imps.py
@@ -11,6 +11,7 @@ from addon_imps.citations import (
 )
 from addon_imps.computing import boa
 from addon_imps.link import dataverse as link_dataverse
+from addon_imps.link import zenodo
 from addon_imps.storage import (
     bitbucket,
     box_dot_com,
@@ -88,6 +89,7 @@ class KnownAddonImps(enum.Enum):
     GITHUB = github.GitHubStorageImp
     GITLAB = gitlab.GitlabStorageImp
     DROPBOX = dropbox.DropboxStorageImp
+    ZENODO = zenodo.ZenodoLinkImp
 
     BOA = boa.BoaComputingImp
 
@@ -112,9 +114,10 @@ class AddonImpNumbers(enum.Enum):
     DATAVERSE = 1010
     GITLAB = 1011
     BITBUCKET = 1012
+    GITHUB = 1013
 
     LINK_DATAVERSE = 1030
-    GITHUB = 1013
+    ZENODO = 1031
 
     BOA = 1020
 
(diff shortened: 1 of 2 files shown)


addon_service/common/known_imps.py
```diff
@@ -88,6 +89,7 @@ class KnownAddonImps(enum.Enum):
     GITHUB = github.GitHubStorageImp
     GITLAB = gitlab.GitlabStorageImp
     DROPBOX = dropbox.DropboxStorageImp
+    ZENODO = zenodo.ZenodoLinkImp
```
cslzchen, line 92: Suggestions:

* Let's make a PR to our feature branch by sorting the the addons first by type and then alphabetically
* Then, make a change to this PR to add `LINK_ZENODO` after `LINK_DATAVERSE`

addon_service/common/known_imps.py
```diff
@@ -112,9 +114,10 @@ class AddonImpNumbers(enum.Enum):
     DATAVERSE = 1010
     GITLAB = 1011
     BITBUCKET = 1012
+    GITHUB = 1013
 
     LINK_DATAVERSE = 1030
-    GITHUB = 1013
+    ZENODO = 1031
```
cslzchen, line 120: A similar case here, let's re-order in the feature PR and then add `LINK_ZENODO` after `LINK_DATAVERSE`.

I am not sure if it's too late the change the number since it may involve DB migration. Can we set STORAGE addons to start with 10xx, CITATION ones with 11xx, COMPUTING ones with 12xx and LINK ones start with 13xx?

cslzchen (changes requested):
In addition to my comments, can we add some comments or Docstring in the implementation file `addon_imps/link/zenodo.py` to info readers of what each class methods do.

Another note, after we feel good with the reference PR, we can close this PR (to get rid of the conversations, JIRA stuff, etc.) and open a new clean PR as the final reference PR.

addon_service/common/known_imps.py
opaduchak: Tiny remark: there's no reason to make `ZENODO` into `LINK_ZENODO`. I had to do this for `Dataverse` because there was already storage implementation for it

addon_service/common/known_imps.py
opaduchak: I concur, but this will definitely require db migration, but this migration should not be very difficult to pull off

cslzchen (commented):
Looks good. Let me make a PR to the feature branch to adjust the order. I won't change the number to avoid migration for now. starting from 103x is good enough for now. If we ever run out of 1001~1029 for other types of addons in the future, we will handle it when it happens.

PR #295, opened 2025-05-27
[ENG-8077] Add `int_supported_features` to dataverse link

cslzchen:
## Purpose

We still need the `int_supported_features` field on the ExternalLinkService model because it is the field FE uses to show the Terms of Service page.

## Changes

* Add `int_supported_features` and related
* Adjust `int_supported_resource_types` and related


addon_service/migrations/0014_authorizedlinkaccount_configuredlinkaddon_and_more.py
```diff
@@ -88,7 +88,7 @@ class Migration(migrations.Migration):
                     models.BigIntegerField(
                         null=True,
                         validators=[
-                            addon_service.external_service.link.models.validate_supported_features
+                            addon_service.external_service.link.models.validate_supported_resource_types
```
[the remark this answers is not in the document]
cslzchen: Good catch, I see what happened ... the "refactor" stuff accidentally modified the migrations file ... it is working locally because this one has already been migrated ... let me revert this

opaduchak (approved):
LGTM, But we still need to finalize list of supported features for Link addons

addon_service/external_service/link/models.py
```diff
@@ -10,19 +15,51 @@
...
+    ADD_UPDATE_FILES_PARTIAL = auto()
+    DELETE_FILES = auto()
+    DELETE_FILES_PARTIAL = auto()
+    FORKING = auto()
+    LOGS = auto()
+    PERMISSIONS = auto()
+    REGISTERING = auto()
+    FILE_VERSIONS = auto()
+    COPY_INTO = auto()
+    DOWNLOAD_AS_ZIP = auto()
```
cslzchen, line 30: @adlius , this my lazy copy from `StorageSupportedFeatures`. I wonder what we actually need for LINK. Is it decided by the CSV product provided?

[verified_link.csv](https://github.com/user-attachments/files/20459142/verified_link.csv)

adlius: Yes. This should be determined by what products thinks is the common list of supported features for link addons. Also note that if there is a partially supported feature, we should add `<FEATURE_NAME>_PARTIAL` to the list, as that's how the FE would know if a feature is only partially supported.

adlius (changes requested):

addon_service/external_service/link/models.py
cslzchen: Thanks for the explanation. I assume `FALSE` means the feature should not be listed. So it should be like this

```python
FORKING_PARTIAL = auto()
LOGS_PARTIAL = auto()
PERMISSIONS = auto()
```

PR #335, opened 2025-10-22
[ENG-8975] Feature/redirect service support

futa-ikeda:
## Ticket

[//]: # (Link to a JIRA ticket if available.)

## Purpose

[//]: # (Briefly describe the purpose of this PR.)

## Changes

[//]: # (Briefly describe or list your changes.)

## Side Effects

[//]: # (Any possible side effects?)

## API Documentation

[//]: # (Any API documentation updates?)

## QE Notes

[//]: # (Any QA testing notes for QE?)

## CE Notes

[//]: # (Any server configuration and deployment notes for CE?)


sh-andriy (approved):
Looks good overall, this follows the existing patterns well and the implementation is clean.

One small cleanup needed in `addon_service/admin/__init__.py:110-112`:

```
enum_multiple_choice_fields = {
      "int_supported_features": StorageSupportedFeatures,
  }
```

`ExternalRedirectService` doesn't have an `int_supported_features` field, so this config does nothing. Should just remove these lines.

Optionally, for consistency with other service types, you could add a `validate_redirect_imp_number` validator in **validators.py** and call it in a `clean()` method on the model - but not blocking since the current setup still works fine.

Otherwise LGTM

futa-ikeda:
> `ExternalRedirectService` doesn't have an `int_supported_features` field, so this config does nothing. Should just remove these lines.
> 
> Optionally, for consistency with other service types, you could add a `validate_redirect_imp_number` validator in **validators.py** and call it in a `clean()` method on the model - but not blocking since the current setup still works fine.

I've opted for just removing those lines. My thinking is, it may be best to keep it as lightweight as possible, and if the need arises, we can add the validator later?

